PR #2, opened 2018-08-14
RFC: dataset document definition

b5:
This is a first draft to get our dataset document definition on the record. It's primarily an assembly of documentation we've written elsewhere

### DIFF
--- a/text/0002-qri_dataset_definition.md
+++ b/text/0002-qri_dataset_definition.md
@@ -337,12 +337,13 @@ a formal alternative.
 # Prior art
 [prior-art]: #prior-art
 
-- [DCAT]
-- [JSON-LD]
-- [OKFN Data Package]
-- [Dat Data Package]
-- [R Packages]
-- [Web Archives]
+- [Git](https://git-scm.com)
+- [DCAT](https://www.w3.org/TR/vocab-dcat/)
+- [JSON-LD](https://en.wikipedia.org/wiki/JSON-LD)
+- [OKFN Data Package](https://frictionlessdata.io/data-packages/)
+- [Dat Data Package](https://datproject.org/)
+- [R Packages](http://r-pkgs.had.co.nz/data.html)
+- [Web Archives WARC spec](https://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.0/)
 
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions


b5:
@rgardaphe, all that time we spent putting together talk notes just made it into an RFC 😄. Mind having a 👀? kthxbyyyyyyyye

text/0002-qri_dataset_definition.md
```diff
@@ -0,0 +1,341 @@
...
+to table.
+
+### Versioned
+Qri’s dataset versioning system is inspired by git, and signs each commit with 
+your identifying keypair. Because qri is only about datasets, qri can do new 
+things like generate commit messages for you.
+
+### Content Addressing
+
+Qri assumes it's underlying store is a content-addressed file system.
```
rgardaphe: should read: Qri assumes its underlying store is....

text/0002-qri_dataset_definition.md
```diff
@@ -0,0 +1,341 @@
...
+things like generate commit messages for you.
+
+### Content Addressing
+
+Qri assumes it's underlying store is a content-addressed file system. 
+All datasets on qri are immutable. Datasets are identified by their 
+cryptographic hash, and assume they are being stored on a content-addressed 
+file system (content is referred to by cryptographic hash). Changes to datasets 
+are stored by creating a new version of data that references the previous 
+version. All Versions of Qri includes a naming systemthat connects
```
rgardaphe: make 'Versions' with a lowercase v.

text/0002-qri_dataset_definition.md
```diff
@@ -0,0 +1,343 @@
...
+
+#### `Citation`
+
+_example:_
+```json
+  { 
+    "name" : "sean carter",
+    "url" : "https://jayz.com",
+    "email" : "hova@jayz.com"
+  }
```
ramfox: I'm confused why we don't have peerids in citations, and why we don't have peernames in users and citations. 

And I guess why citations are a slice of citations rather than a slice of users.

I can see the logic that a citation may not be a user (peer), but having a way to give credit that will link to another peer will probably be useful in the future, when people work on teams or folks are referencing other qri peers when making datasets.

text/0002-qri_dataset_definition.md
```diff
@@ -0,0 +1,343 @@
...
+body that lists images leaves ambiguity around how it would be stored and 
+represented. Orienting around a document model does not alleviate the burden
+package managers take on more directly.
+
+#### Graph vs. File
+One of the key tensions is between _Graph_ and _File_ representations. By
+choosing to orient around a file metaphor instead of a database metaphor,
+it's harder to see how this document model has value in relation to DAG-based
+storage systems.
+
```
ramfox: Are you saying that 

"I understand that since we are in a DAG based storage system, it would seem more logical to orient around the metaphor of datasets as graphs rather than documents"?

If so, maybe we need to add explicitly somewhere why exactly we orient around a file rather than a graph. I understand that basically this entire rfc is an argument for exactly that, this section or a previous one might benefit from a little more clarity on that point.

text/0002-qri_dataset_definition.md
b5: > I'm confused why we don't have peerids in citations, and why we don't have peernames in users and citations.

We actually do have peerID's in Citations, in an ID field that isn't identified here. I'll update.

> And I guess also confused why citations are a slice of citations rather than a slice of users.

Citations can be both a _thing_ and a _person_. One can cite a person saying something, or they can quote a book or text (with a URL) 

In both these cases, we're working within the context of an existing Metadata specification (DCAT & the Dublin Core) and I want to stay as close to it as possible so we get better metadata interop. So it means that we're kinda stuck shoehorning our own user info into an existing pattern.

text/0002-qri_dataset_definition.md
```diff
@@ -0,0 +1,352 @@
...
+# Prior art
+[prior-art]: #prior-art
+
+- [DCAT]
+- [JSON-LD]
+- [OKFN Data Package]
+- [Dat Data Package]
+- [R Packages]
+- [Web Archives]
+
```
ramfox: Should these link to somewhere?

b5: lol yes they should. Will do now.

PR #4, opened 2018-08-14
RFC: structured data io

b5:


text/0004-structured_data_io.md
```diff
@@ -0,0 +1,215 @@
...
+which are elements of a composite data structure such as an array or object.
+Structured reader & writer interfaces combine a byte stream, data format 
+and schema to create entry readers & writers that produce & consume entries
+instead of bytes. Structured I/O streams can be composed & connected to form
+the basis of rich data communication capable of spanning across formats.
+
+# Motivation
+[motivation]: #motivation
+
+Structured I/O is the primary means of _doing things with data_. All of the
```
osterbit: Obviously Structured I/O is about input and output but I misread this sentence and expected that it was also intended to provide higher level facilities for "doing things with data".  not that it needs to be changed but just flagging that

b5: Great point. "Doing things with data" is getting ahead of ourselves, and positions this API as being high-level, when it's in fact a lower-level parsing API. That has me thinking about a higher-level API? We've talked about a `stats` package & statistical metadata for a while, maybe it's time to think about an RFC for a package based around structured I/O primitives that holds common statistical functions?

text/0004-structured_data_io.md
```diff
@@ -0,0 +1,215 @@
...
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Given that we've already written this, the time for considering alternatives
+should be in future a RFC.
+
+# Prior art
+[prior-art]: #prior-art
+
+[golang's io package](https://godoc.org/io) is _the_ source of inspiration here.
+
```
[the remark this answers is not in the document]
b5: I don't think something like csvutil does, given that it's predicated on a single data format. the purpose of structured I/O is to create a stream of parsed data, and deliver a set of expectations about that parsed data at the same time.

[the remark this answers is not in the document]
osterbit: so the additional benefit this package provides is combined json and csv parsing? I still think comparisons to adjacent options are both helpful in contextualizing the functionality and justifying the necessity and design of a set of interfaces like this.

text/0004-structured_data_io.md
```diff
@@ -12,33 +11,49 @@ we'd hope, or less complete than we'd expect from a new RFC.
...
 [summary]: #summary
 
-Structured I/O defines interfaces for reading & writing streams of _entries_,
-which are elements of a composite data structure such as an array or object.
-Structured reader & writer interfaces combine a byte stream, data format 
-and schema to create entry readers & writers that produce & consume entries
-instead of bytes. Structured I/O streams can be composed & connected to form
-the basis of rich data communication capable of spanning across formats.
+Structured I/O defines interfaces for reading & writing streams of parsed 
+ _entries_, which are elements of dynamic-yet-structred data such as JSON, CBOR,
```
dustmop: Type: structred -> structured

Also, by "dynamic" I believe you mean "arbitrary"? Or, next paragraph, you use "generic-yet-structured".

text/0004-structured_data_io.md
```diff
@@ -1,26 +1,52 @@
...
+dataset. Structured I/O is intended to be a robust set of primitives for
+performing these tasks. Structured _streams_ (readers & writers) abstract away 
+the data formats, automatically parsing raw bytes into parsed _entries_ of data.
+Working with entries instead of bytes allows the programmer to avoid thinking
+about the underlying format & focus on the semantics.
+
+Orienting our primitives around _streams_ helps manage concerns created by both 
+network latency and data volume. Stream-based programming reduces the
+
+Structured I/O builds on foundations set fourth in the _structure_ portion of
```
dustmop: Typo: "set forth"

text/0004-structured_data_io.md
```diff
@@ -83,15 +98,30 @@ reader from scratch & reading it's values:
...
 
   _, err := str.ReadEntry()
   fmt.Println(err.Error()) // EOF
 ```
 
-### Stream
+### Stream & Top Level Data
 A _stream_ refers to refer to both a _reader_ and a _writer_ collectively.
 
+_Top Level_ refers to the first entry in a discrete set of data.
```
dustmop: This is not totally accurate. "Top-level" refers to the specific type of the outermost element in a structured piece of data. It's an orthogonal concept to "entries".

And while I know it's repeated below, I think it would help to specify in this section that Qri requires a top-level type of either array or object.

text/0004-structured_data_io.md
```diff
@@ -100,6 +130,51 @@ Entries are themselves a small abstraction that carries the `Value` (parsed
 data), `Index` and `Key`. Only one of `Index` and `Key` will be populated at a
 given point, depending on weather an array or object is being read.
```
dustmop: Add something to the effect of "`Value` is of an arbitrary structure", to cut off any expectations that it is merely a primitive type.

b5:
> if my reading of this as "io functionally similar to protocol buffers with an improved interface" is accurate referencing them might be helpful

That isn't accurate, but the statement is helpful! I now realize that I haven't taken enough time to lay out the problem space we're operating on. I've added more details to the RFC motivation section that hopefully better articulates what we're working with.

But to speak to Protocol Buffers specifically, PB rely on two key assumptions that are not present in this context:
* schemas are defined in advance, at compile time
* the data being passed around will match that schema

Neither of those are true for us. Our schemas are defined _at runtime_. _And_ we need to be able to pass around & work with data that doesn't match the provided schema. Because of that Protocol Buffers are not comparable to what's going on here, because they start with assumptions we cannot make.

For that reason I want to avoid describing things in terms of static type serialization, and instead focus on dynamic examples. I've added a reference to OpenAPI as what I believe is a more accurate kind of Prior art we're building off of.

Does that help at all? I want to make sure we add as much detail as possible to get the need for these primitives across.

b5:
well then in it goes 🎉

text/0004-structured_data_io.md
```diff
@@ -12,33 +11,49 @@ we'd hope, or less complete than we'd expect from a new RFC.
...
+spanning across formats.
 
 # Motivation
 [motivation]: #motivation
+
+One of the prime goals of qri is to to be able to make any dataset comparable to
+another dataset. Datasets are also intended to be a generic-yet-structured
+document format, able to support all sorts of data with varying degrees of
+quality. These requirements mean datasets must be able to define their own 
+schemas, and may include data that violates that schema.
```
osterbit: this may be a dumb question but, what does a schema tell me about the data if it's the case that it may or may not be invalid? (seeing as we have access to previous schemas, that were valid).  In a typical world where a schema can't be invalid, I interpret the schema as the source of truth.  If a change invalidates it, and let's say removes the schema or rebuilds it from the data dynamically, I can both go back to the old version to see what the truth used to be and look at the current (possibly absent) schema and know what the truth is now.  But if the schema may or may not be invalid I'm not sure what I'm supposed to do with that.

b5: No a dumb question at all!

In the case where data may or may not be valid, the schema is a pointer to the _authorial intent_ of the dataset. It's the equivalent of "this is how the data _should_ look". In this context, the schema isn't a source of truth, it's a source of _intent_ that is only a source of _truth_ if all data matches, or we're operating in "strict mode".  

That "should look like" can be very useful in different context for higher-level decision making. For example, the schema may dictate that a field labeled "date" should be a datestamp in the format `YYYY-MM-DD`. I can use the schema to make smarter decisions about how to coerce invalid data. At this level of API design, however, "smarter decisions" should be kicked up the chain for different implementations to worry about, because one size doesn't fit all, and it'd be too tough to know how to solve those specific problems in generalized ways. So instead we focus on an interface that makes sure the schema is on hand should you want to use it in a higher-level API. We don't require you to make use of it at all if it doesn't help a given use case.

Finally, it's not always true that we have access to previous schemas that were valid.  Peers don't always have all of a dataset history on hand. Even when we _do_ have the full history on hand, there is no guarantee that a completely valid schema ever existed in the dataset history, which is why we fall back to identity schemas instead of getting entangled with previous maybe-or-maybe-not-fully-valid schemas, because one of the two Identity schemas is _required_ to be valid to create a dataset in the first place.

PR #6, opened 2018-08-14
RFC: dataset naming

b5:


text/0006-dataset_naming.md
```diff
@@ -0,0 +1,97 @@
...
+as follows:
+
+    dataset_reference = peer_name/dataset_name@/network/hash
+
+an example of that looks like this:
+
+    b5/comics@QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y/ipfs/QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y
+
+In a sentence: b5 is the name of a peer, who has a dataset named comics, and 
+it’s hash on the ipfs network at a point in time was `QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y`
```
rgardaphe: should read, "and its hash on the ipfs network"

text/0006-dataset_naming.md
```diff
@@ -0,0 +1,97 @@
...
+me/comics
+
+In a sentence: “the latest version of my dataset named comics.” Under the hood, we’ll re-write your request to your peername for you.
+drop names with hashes
+
+Finally, it’s also possible to use just the hash. This is a perfectly valid dataset reference:
+
+/ipfs/QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y
+
+In this case we’re ignoring naming altogether and simply referring to a dataset by it’s network and hash. Because IPFS hashes are global, we can do this across the entire network. If you’re coming from git, this is a fun new trick.
```
rgardaphe: should read: "referring to a dataset by its network and hash"

text/0006-dataset_naming.md
```diff
@@ -0,0 +1,97 @@
...
+We need peer names so lots of people can name datasets the same thing, and in 
+this instance that giant hash thing is used to refer to a dataset at a specific 
+version of a dataset, from an exact point in time.
+default to latest on ipfs
+
+Now, having to type b5/comics@/ipfs/QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y 
+every time you wanted a dataset would be irritating. So we have two defaults. 
+The default network is ipfs, and the default hash is the lastest known version 
+of a dataset. We say latest known because sometimes things can fall out of sync. 
+If you’r only working with your own local datasets, this won’t be an issue.
```
rgardaphe: should read, "If you're only working with"

text/0006-dataset_naming.md
```diff
@@ -0,0 +1,97 @@
...
+
+# Summary
+[summary]: #summary
+
+Define the qri naming system and conventions for name resolution.
+
+# Motivation
+[motivation]: #motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
```
dustmop: This is still the placeholder text.

How about:

"Qri uses a number of jargon terms around dataset names that need clarification and consistent usage."

text/0006-dataset_naming.md
```diff
@@ -0,0 +1,97 @@
...
+
+    b5/comics@QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y/ipfs/QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y
+
+In a sentence: b5 is the name of a peer, who has a dataset named comics, and 
+it’s hash on the ipfs network at a point in time was `QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y`
+
+We need peer names so lots of people can name datasets the same thing, and in 
+this instance that giant hash thing is used to refer to a dataset at a specific 
+version of a dataset, from an exact point in time.
+default to latest on ipfs
```
dustmop: Is this a new section? Needs markdown to headerify it.

text/0006-dataset_naming.md
```diff
@@ -0,0 +1,97 @@
...
+of a dataset. We say latest known because sometimes things can fall out of sync. 
+If you’r only working with your own local datasets, this won’t be an issue.
+
+Anyway, that means we can cut down on the typing if we just want the latest 
+version of b5’s comics dataset, we can just type:
+
+b5/comics
+
+In a sentence: “the latest version of peers b5’s dataset named comics.”
+the me keyword
```
dustmop: Headerification markdown.

text/0006-dataset_naming.md
```diff
@@ -0,0 +1,97 @@
...
+/ipfs/QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y
+
+In this case we’re ignoring naming altogether and simply referring to a dataset by it’s network and hash. Because IPFS hashes are global, we can do this across the entire network. If you’re coming from git, this is a fun new trick.
+To recap:
+
+All of these are ways to refer to a dataset:
+
+    peer_name/dataset_name : b5/comics
+    network/hash : /ipfs/QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y
+    peer_name/dataset_name@/network/hash: b5/comics@/ipfs/QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y
```
dustmop: We probably need names for what these different kinds of references are. The code already does this in some informal ways, but the code is confusing because the different kinds not explicitly mentioned. Some possibilities: "user-friendly reference", "network-hash reference", "canonical reference".

text/0006-dataset_naming.md
```diff
@@ -0,0 +1,97 @@
...
+are to make sense of things. The full definition of a dataset reference is 
+as follows:
+
+    dataset_reference = peer_name/dataset_name@/network/hash
+
+an example of that looks like this:
+
+    b5/comics@QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y/ipfs/QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y
+
+In a sentence: b5 is the name of a peer, who has a dataset named comics, and
```
dustmop: Perhaps this belongs in a different RFC, but I would expect / appreciate having definitions here giving exact terms for the sections of this dataset_reference. I say perhaps because I'm not sure we've landed on the exact terms to use everywhere.

For example:
peer_name = b5, the name of the peer that created this dataset
...

Also, the hash Qme...Me2y appears twice with no mention of why.

ramfox: I propose:

peername/dataset_name@peer_id/network/dataset_id

As names.

Could also go with peer_hash and dataset_hash, or id and hash, respectively.

The hash Qme...Me2y appearing twice is a mistake, the peer ID and dataset ID should be different.

b5: I'm a fan of peer_id and dataset_id. This layer should theoretically be content-addressing agnostic. dataset naming should rely on unique IDs, which is one of a few properties of content-addressing  / identify-by-hash systems.

text/0006-dataset_naming.md
b5: This RFC is supposedly putting fourth the concept of dataset naming in the first place. I'd propose:

> As a decentralized system, qri must confront the [Zooko's Triangle problem](https://en.wikipedia.org/wiki/Zooko%27s_triangle) which is establishing a way of referring to datasets that is:
> - human readable
> - decentralized
> - unique
> Because Qri is assumed to be built atop a content-address file system as it's storage layer, the properties of being decentralized & unique are already present.
> The Qri naming system maps a human readable name to the _tip_ of a dataset history.

text/0006-dataset_naming.md
dustmop: Any feelings about referring to the final component as "version_id", since it's specifying the exact version of the dataset? All the rest of the names are good to me, I just think it would be too confusing to have both "dataset_name" and "dataset_id".

Right now, the code in repo/ref is using ProfileID to refer to the component after the '@' sign but before the network protocol. I'm fine with changing this to PeerID instead. With peers, there's a direct one-to-one correlation between peername and peerID, whereas datasets using a one-many relationship between dataset_name and version_ids.

b5:
This looks great to me! As the author I can't approve, but would love it if others could have a quick look. Let's push the merge button

ramfox:
V fun.

I like the change from `peername` to `handle`!!!

One thing, a few locations use the term `user`, and previously we have been avoiding the term `user` in favor of `peer`. Specifically in lines 39, 45, and 127 (I'm not including the phrase 'user-friendly', as that does not relate to the concept of `peer` that we use in the codebase and docs. Any objection to changing those over, so they match other docs?

b5:
Yeah might as well not introduce more language. I'll update `user` -> `peer`...

ramfox:
💅 💇 🥇

PR #3, opened 2018-08-14
RFC: content addressed file system

b5:
first draft to get cafs into the rfc process.

PR #7, opened 2018-08-14
RFC: repo

b5:
This RFC is a great chance for us to work to add a little clarity to the role & interface of repo, especially as it relates to #3.

I'm open to renaming repo here if we decide that's appropriate.

### DIFF
--- a/text/0007-repo.md
+++ b/text/0007-repo.md
@@ -1,0 +1,81 @@
+- Feature Name: repo
+- Start Date: 
+- RFC PR: (leave this empty)
+- Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+repo defines an interface for local storage & retrieval of qri data.
+
+# Motivation
+[motivation]: #motivation
+
+TODO
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+TODO
+
+# Reference-level explanation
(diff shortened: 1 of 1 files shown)


text/0007-repo.md
```diff
@@ -0,0 +1,81 @@
...
+
+  // A repository must maintain profile information about the owner of this dataset.
+  // The value returned by Profile() should represent the peer.
+  Profile() (*profile.Profile, error)
+  // SetProfile sets this repo's current profile. Profiles must contain a private key
+  SetProfile(*profile.Profile) error
+  // PrivateKey hands over this repo's private key
+  PrivateKey() crypto.PrivKey
+  // A repository must maintain profile information about encountered peers.
+  // Decsisions regarding retentaion of peers is left to the the implementation
```
dustmop: "Decisions"

If we can change this interface, I am not at all a fan of having two methods with completely different semantics that only vary in name by a single letter. Profile represents the peer's own info, while Profiles is other peers we've seen in the wild, which is different in more important ways than simply being plural. I would rename it to something like "RemoteProfiles" or "KnownProfiles".

ramfox: Throwing "PeerProfiles" into the ring.

b5:
Closing. This RFC is stale

PR #13, opened 2018-08-14
RFC: html viz

b5:


b5:
Hey team, I'm picking this PR back up b/c we never properly finished it. I've added a note explaining the time-warp. Would love a review.

text/0011-html_viz.md
```diff
@@ -0,0 +1,177 @@
...
+
+Having default empty pointers prevents unnecessary `if` clauses, allowing a skip to tests for existence on component fields:
+```html
+{{ if ds.transform.script }}
+{{ end }}
+```
+
+### Template functions
+
+Top level functions should be loaded into the template `funcs` space to make rendering templates easier. The go html template package comes with [predefined functions](https://golang.org/pkg/text/template/#hdr-Functions), all of which are included in this RFC. An example prints the length of the body:
```
ramfox: I don't totally understand what you mean by "all of which are included in this RFC"

Do you mean that there are examples of all of those functions someone in this RFC, or that this RFC proposes adding them all into our templating?

b5: Good point, I meant the latter. That this RFC will de-facto introduce all of these functions into our template engine unless we explicitly turn them off (because we're using the template package), and I don't think we should turn them off.

text/0011-html_viz.md
```diff
@@ -0,0 +1,182 @@
...
+
+Qri has a syntax-agnostic `viz` component that encapsulates the details required to visualize a dataset. This RFC proposes the first & default visualization syntax should be rendering to a single HTML document using a template processing engine.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+
+### `qri render` & default template
+
+A command will be added to both Qri's HTTP API ('API' for short) & command-line interface (CLI) called `render`, which adds the capicty to execute HTML templates against datasets. When called with a specified dataset `qri render` will load the dataset, assume the viz syntax is HTML, and use a default template to write an HTML representation of a dataset to `stdout` on the CLI, or the HTTP response body on the API:
```
dustmop: capicty -> capacity

text/0011-html_viz.md
```diff
@@ -0,0 +1,182 @@
...
+
+The default template and output path can be overridden with the `--template` and `--output` flags respectively. The output is on the CLI only:
+`qri render --template template.html --output rendered.html me/example_dataset`
+
+The default template must be carefully composed to balance the size of the resulting HTML file in bytes against readability & utility of the resulting visualization. It should also include a well-constructed citation footer that details the components of a dataset in a concise, unobtrusive manner that invites users to audit the veracity of the dataset in question. These defaults should encourage easy reading and invite verification on the part of the reader.
+
+### Vizualizations in datasets
+
+Saving a dataset will by default execute the default template to a file called `index.html` & place it in the IPFS merkle-DAG. When an IPFS HTTP gateway receives a request for a DAG that is a directory containing `index.html`, it returns the HTML file by default. This means when a user visits a dataset on the d.web–completely outside the Qri system of dataset production–they are greeted with a well-formatted dataset document by default.
```
dustmop: This is super clever. Do we know for sure that IPFS gateways serve index.html instead of showing a directory listing? I thought that might be an Apache specific thing. Is there a way to force the IPFS gateway to show the datasets contents?

text/0011-html_viz.md
```diff
@@ -0,0 +1,182 @@
...
+While care will be taken to keep `index.html` files small, users may understandably want to disable them entirely. To achieve this we'll add a new flag to `qri save` and `qri update`: `--no-render`. No render will prevent the execution of any viz template. This will save a few KB from version to version at the cost of usability.
+
+Users can _override_ the default template by supplying their own custom viz templates either by specifying a `viz.scriptPath`:
+
+`dataset.yaml`:
+```yaml
+name: example_dataset
+# additional fields elided ...
+viz:
+  syntax: html
```
dustmop: This should be "format": "html"

text/0011-html_viz.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Template API
+
+Introducing HTML template execution requires defining an API for template values. This API will need to be documented & maintained just like any other API in the Qri ecosystem.
+
+The template implementation will use the [go language html template package](https://golang.org/pkg/html/template)
+
+#### Dataset is exposed as `ds`
+
+HTML template should expose a dataset document at `ds`. By exposing the document as `ds`, it matches our for referring to a dataset in starlark., and allows access to all defined parts of a dataset. ds should use _lowercase_ fields for component accessors. eg:
```
dustmop: "it matches our for" is missing a word

text/0011-html_viz.md
```diff
@@ -0,0 +1,182 @@
...
+
+Having default empty pointers prevents unnecessary `if` clauses, allowing a skip to tests for existence on component fields:
+```html
+{{ if ds.transform.script }}
+{{ end }}
+```
+
+### Template functions
+
+Top level functions should be loaded into the template `funcs` space to make rendering templates easier. The go html template package comes with [predefined functions](https://golang.org/pkg/text/template/#hdr-Functions). Because our implementation builds the html/template package, this RFC introduces all of these functions into our template engine. I think this is totally fine, might be a pain if someone needs to write a non-go `Render` implementation, but at least the .
```
dustmop: "Because our implementation builds the" -> "Because our implementation builds *on* the"

Sentence is incomplete at the end of this paragraph

text/0011-html_viz.md
```diff
@@ -0,0 +1,182 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+One of the biggest drawbacks to the proposed implementation comes from a natural tendency to use the following to template the contents of a dataset body in to a dataset:
+
+```html
+<script type="text/javascript"> window.data = {{ ds.get_body }} </script>
+```
+
+Using templating this way prints the entire body into the template. If the body is large, this will cause _all body data to be duplicated in `index.html`_. I propose we fight this through documentation, because there are many scenarios where this code is completely appropriate (for example, when the dataset body is an aggregation of some other data source).
```
dustmop: I'm not sure we should even support this scenario. If someone wants to access the entirety of the body, perhaps we should force them to explicitly copy the body into the stats somewhere. That is, by requiring additional, explicit steps, instead of providing access by default, we would imply that it's not a recommended course of action.

text/0011-html_viz.md
b5: It's the default gateway behaviour for unixfs v1 dags. Looking at the code, I don't think there's a way to force listing directory contents: https://github.com/ipfs/go-ipfs/blob/74d07eff35965a3f635d03aedaa43561c73679e2/core/corehttp/gateway_handler.go#L249-L274

That seems like it would be useful. Might make for a nice PR into go-ipfs to add a query param, I have no idea if they'd accept. @lgierth might be a good person to ask 😄.

In the meantime tho, our `citation` partial should list a relative link to the `/dataset.json` definition, which should help a bit.

dustmop: Yeah, linking to the dataset.json file should work fine, since it functions as a self-describing doc of the dataset.

text/0011-html_viz.md
b5: Yeah, my concern is there are situations where it absolutely _is_ the recommended course of action. For instance, [b5/country_codes](https://app.qri.io/b5/country_codes) only has 250 entries, If for some reason we needed to graph those, It would be fine to embed the entire body.

b5:
neato gang. this club is formed

PR #14, opened 2018-08-14
RFC: CLI commands

ramfox:
List of commands, flags, and expected behavior

### DIFF
--- a/text/0012-CLI-commands.md
+++ b/text/0012-CLI-commands.md
@@ -835,9 +835,10 @@ commands to control how your datasets are published to registries, opting out
 on a dataset-by-dataset basis.
 
 By default qri is configured to publish to https://registry.qri.io,
-the main public collection of datasets & peers. "qri add" and "qri update"
+the main public collection of datasets & peers. "qri new" and "qri update"
 default to publishing to a registry as part of dataset creation unless run 
-with the "no-registry" flag.
+with the "no-publish" flag. Published datasets that are under 20MBs will be
+backed up by the Qri registry.
 
 Unpublished dataset info will be held locally so you can still interact
 with it. And your datasets will be available to others peers when you run 
--- a/text/0012-CLI-commands.md
+++ b/text/0012-CLI-commands.md
@@ -587,7 +587,7 @@ Create a new dataset
 
 
 New creates a dataset from data you supply. Please note that all data added
-to qri is made public on the distributed web when you run qri connect
+to Qri is made public on the distributed web when you run qri connect
 
 When adding data, you can supply metadata and dataset structure, but it’s not
 required. qri does what it can to infer the details you don’t provide.
@@ -597,8 +597,13 @@ qri currently supports three data formats:
 - CBOR (Concise Binary Object Representation)
 
 Once you’ve added data, you can use the export command to pull the data out of
-qri, change the data outside of qri, and use the save command to record those
-changes to qri.
+Qri, change the data outside of Qri, and use the save command to record those
+changes to Qri.
+
+Datasets added to Qri using `new` are published to the registry by default.
+Use the `--no-publish` flag to not publish this dataset to the registry. For
+more information on the registry and why it's important check out the [registry
+section](#qri_registry).
 
 ```
 qri new [flags]
@@ -622,7 +627,7 @@ create a dataset with a dataset data file:
   -h, --help              help for new
   -m, --message string    commit message
       --private           make dataset private. WARNING: not yet implimented. Please refer to https://github.com/qri-io/qri/issues/291 for updates
-  -p, --publish           publish this dataset to the registry
+  -p, --no-publish        do not publish this dataset to the registry
       --secrets strings   transform secrets as comma separated key,value,key,value,... sequence
   -t, --title string      commit title
 ```
--- a/text/0012-CLI-commands.md
+++ b/text/0012-CLI-commands.md
@@ -50,6 +50,22 @@ The Qri CLI commands are as follows:
 * [qri validate](#qri_validate)  - Show schema validation errors
 * [qri version](#qri_version)  - Print the version number
 
+We are using the cobra package to create each command.
+
+Each command has a struct `[name_of_command]Options` (eg `ValidateOptions`)
+
+Each option struct requires `Complete`, `Validate`, and `Run` methods.
+
+The `Complete` method fills in any parameters needed for the command to run. It 
+also adds the command arguments to the option struct.
+
+The `Validate` methods ensures that the parameters added to the option struct
+are all valid, and returns an error describing the problem if the options are
+invalid.
+
+The `Run` method calls to the more core Qri functions to run the command using
+the parameters in the options struct.
+
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
@@ -428,28 +444,46 @@ qri get [flags] <dataset_reference> <dataset_reference2> ...
 <a id='qri_info'></a>
 ## qri info
 
-show summarized description of a dataset
+Show summarized description of a dataset
 
 ### Synopsis
 
-info describes datasets
+Info describes datasets. By default, it will return the peername, dataset name, 
+the network, the dataset hash, the file size, the length of the datasets, 
+and the validation errors.
+
+Using the `--format` flag, you can get output in json. This will return a json
+representation of the dataset, without the dataset body, identical to 
+`qri get --format json`.
+
+To get info on a peer's dataset, you must be running `qri connect` in a separate 
+terminal window.
 
 ```
-qri info [flags]
+qri info [flags] <dataset_reference>
 ```
 
 ### Examples
 
 ```
-  get info for b5/comics:
-  $ qri info b5/comics
+  # get info for my dataset:
+  qri info me/annual_pop
 
-  get info for a dataset at a specific version:
-  $ qri info me@/ipfs/QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn
+  # get info for a dataset at a specific version:
+  qri info me@/ipfs/QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn
 
   or
 
-  $ qri info me/comics@/ipfs/QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn
+  qri info me/comics@/ipfs/QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn
+
+  # get info in json format
+  qri info -f json me/annual_pop
+
+  # to get info on a peer's dataset, spin up your qri node
+  qri connect
+
+  # then, in a separate window, request the info from peer b5
+  qri info b5/comics
 ```
 
 ### Options
@@ -464,15 +498,18 @@ qri info [flags]
 <a id='qri_list'></a>
 ## qri list
 
-show a list of datasets
+Show a list of datasets
 
 ### Synopsis
 
 
-list shows lists of datasets, including names and current hashes. 
+List shows lists of datasets, including names and current hashes. 
 
 The default list is the latest version of all datasets you have on your local 
 qri repository.
+
+When used in conjuction with `qri connect`, list can list a peer's dataset. You
+must have `qri connect` running in a separate terminal window.
 
 ```
 qri list [flags]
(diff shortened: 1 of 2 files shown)


text/0012-CLI-commands.md
```diff
@@ -0,0 +1,462 @@
...
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+<!-- - Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this? -->
+
+# Prior art
+[prior-art]: #prior-art
```
b5: I think it'd be great to list kubernetes here & link to the github repo (possibly even link to the package that includes the `NewKubernetesCommand` stuff?)

text/0012-CLI-commands.md
```diff
@@ -0,0 +1,462 @@
...
+```
+      --concise         print output without indentation, only applies to json format
+  -f, --format string   set output format [json, yaml] (default "yaml")
+  -h, --help            help for get
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
```
b5: For this section we could say:
> Structuring qri around a command line client as our base implementation will mean that CLI features routinely outpace other ways of interacting with qri (like the API and subsequently the frontend). To mitigate this it's important that we view any proposed feature from perspectives other than just the CLI, and work to maintain parity across our outward facing interfaces wherever possible & reasonable.

ramfox:
@b5 plz halp

I'm having trouble with `qri peers disconnect`. Is this meant to be a way to close your connection to a peer, so that the peer is essentially blocked? Or just close your explicit connection to it right now. Will this prevent automatically connecting to this peer in the future?  Is this supposed to be temporary? If you close the connection, can the peer just re-open it using `qri peers connect`? It seems right now that you don't have to be connected to a peer in order for that peer to be connected to you.

It seems like this is an error in libp2p. But regardless, trying to understand the spirit of what `qri peers disconnect` is supposed to be for. I'm worried it can read as a security function, but I'm pretty sure that's not what it's supposed to do.

text/0012-CLI-commands.md
```diff
@@ -0,0 +1,1361 @@
...
+
+body
+diff
+export
+get
+info
+log
+render
+validate
+
```
ramfox: I am asserting this to be true. I believe it was the original goal. Currently, `use` only works correctly with get.

text/0012-CLI-commands.md
```diff
@@ -0,0 +1,1366 @@
...
+Once you’ve added data, you can use the export command to pull the data out of
+Qri, change the data outside of Qri, and use the save command to record those
+changes to Qri.
+
+Datasets added to Qri using `new` are published to the registry by default.
+Use the `--no-publish` flag to not publish this dataset to the registry. For
+more information on the registry and why it's important check out the [registry
+section](#qri_registry).
+
```
ramfox: Again, just deciding this. But very open to changing my mind, as I can see pros and cons for each.

pros: fill out the registry, have more stuff for folks to search, builds the network
cons: annoy or upset users that didn't realize their stuff was searchable by others

Although: Qri is public. It almost seems disingenuous to make people feel like by not 'publishing' their data, it is somehow shielded from the public eye. 

I have so many thoughts about this.

text/0012-CLI-commands.md
```diff
@@ -0,0 +1,1367 @@
...
+can find them through search tools and via web links. You can use registry 
+commands to control how your datasets are published to registries, opting out
+on a dataset-by-dataset basis.
+
+By default qri is configured to publish to https://registry.qri.io,
+the main public collection of datasets & peers. "qri new" and "qri update"
+default to publishing to a registry as part of dataset creation unless run 
+with the "no-publish" flag. Published datasets that are under 20MBs will be
+backed up by the Qri registry.
+
```
ramfox: We've talked about backing up published datasets under a certain size before, and I just picked an arbitrary number.

Not married to leaving this in.

text/0012-CLI-commands.md
```diff
@@ -0,0 +1,1367 @@
...
+storing the data will need to garbage-collect that data when it’s good & ready, 
+which could be anytime. If you’re running low on space, garbage collection will 
+be sooner. 
+
+Keep in mind that by default your IPFS repo is capped at 10GB in size, if you
+adjust this cap using IPFS, Qri will respect it.
+
+Use the `--free` flag to force immediate removal of a dataset from both qri & 
+IPFS to free up the space that was taken up by that dataset.
+
```
ramfox: The docs currently have a line saying "we promise we will add this feature in". I just re phrased and named the flag `free`

text/0012-CLI-commands.md
```diff
@@ -0,0 +1,1367 @@
...
+Remove gets rid of a dataset from your Qri node. After running remove, Qri will 
+no longer list your dataset as being available locally. By default, remove frees
+up the space taken up by the dataset, but not right away. The IPFS repo that’s 
+storing the data will need to garbage-collect that data when it’s good & ready, 
+which could be anytime. If you’re running low on space, garbage collection will 
+be sooner. 
+
+Keep in mind that by default your IPFS repo is capped at 10GB in size, if you
+adjust this cap using IPFS, Qri will respect it.
+
```
ramfox: Am I wrong in sussing that this is an IPFS type command. Can we give guidance on how to change it? Or a way to change this using the Qri config?

text/0012-CLI-commands.md
```diff
@@ -0,0 +1,1367 @@
...
+### Options
+
+```
+      --body string       path to file or url of data to add as dataset contents
+  -f, --file string       dataset data file (yaml or json)
+  -h, --help              help for save
+  -m, --message string    commit message for save
+  -p, --no-publish        do not publish this dataset to the registry
+      --secrets strings   transform secrets as comma separated key,value,key,value,... sequence
```
ramfox: Same publish/no-publish debate. Just highlighting for consistency.

ramfox:
Would love feedback/debate
@b5 @dustmop @osterbit @rgardaphe

text/0012-CLI-commands.md
b5: We should def create documentation on how to change it as part of a larger "configuring IPFS" documentation, and reference it again in another set of docs on "managing Qri Data". I'll open issues over on the website repo to create these.

text/0012-CLI-commands.md
b5: `free` makes me think that this delete somehow costs nothing, or liberates me in some way. How about: 
`--delete force data to be immediately removed`
Doc note could be the same:
```
Use the `--delete` flag to force immediate removal of a dataset from both qri & 
IPFS to free up the space that was taken up by that dataset.
```

text/0012-CLI-commands.md
b5: Should be the opposite of what it is here. `--publish : publish this dataset to the registry`

text/0012-CLI-commands.md
b5: ❤️ . I think we should leave it here for now, and in the future consider another RFC that adds the dataset editing commands to `use`, but requires only one dataset be selected:
- save
- remove
- rename

text/0012-CLI-commands.md
```diff
@@ -0,0 +1,1361 @@
+- Feature Name: CLI commands
+- Start Date: 2018-08-13
+- RFC PR: <!-- (leave this empty) -->
```
b5: change to:
- RFC PR: [#14](https:github.com/qri-io/rfcs/pull/14)

text/0012-CLI-commands.md
```diff
@@ -0,0 +1,1361 @@
...
+- Issue: <!-- (leave this empty) -->
```
b5: remove `- Issue:`, add `- Repo: https://github.com/qri-io/qri`

text/0012-CLI-commands.md
ramfox: So, in the registry docs, it says we publish by default. In the code we don't publish by default.

Whatever it is should be consistent with --new and --save

text/0012-CLI-commands.md
b5: so much this. I think my current feelings on the subject are here: https://github.com/qri-io/qri/issues/513#issuecomment-412281518

> After playing with this, I don't think qri new should publish by default. I know initially we'd agreed on the opposite, but after spending time using qri I've come to two major reasons for this:
> A lot of the datasets I make / work on are "practice" or unfinished work I don't want to announce to the world (yet).
> I often work without an internet connection, which creates inconsistent behaviour in regard to publishing
> In the offline example, qri publish running secretly in the background will fail in this context, which undermines the one upside of "always publish", which is being able to say "I ran new, it should be online". For these reasons, I think we should avoid publishing by default, and instead focus on building up a whole set of tutorials & flows around publication that's similar to git push.

This RFC process has highlighted two things that I think need more thought & clarification:
1. Getting registries right is going to be _very hard_. I think they should be as thin as possible
2. Our story around "remote control" should intersect with this `qri publish` conversation. Eventually we'll need something that's the rough equivalent of `git push`. (For example, an office has a qri node that's always online that you can "push to" so data gets copied there), which feels a lot like `qri publish`.

These are all questions that I'd like to resolve before putting `qri publish` in as a default behavior.

ramfox: oooo, great!! thank you.

text/0012-CLI-commands.md
b5: Yeah the more I've thought about this, the more I think we need the concept of a "dataset preview" that is the head of a dataset, plus a sampling of the body (the hard part is getting "sampling" right). I'd like to get into a position where we can say the registry will _always_ accept the preview of a dataset, and if the registry likes you enough, it'll take the whole thing. For that we'd need _another_ RFC that outlines what "likes you" means, which is the concept of reputation.

So, we have a bunch of work to do. For now I think we should just say "a preview of the dataset", and leave it as a murky thing, because right now it's a murky thing.

b5: could also just remove it 🤷‍♂️

ramfox: I think maybe we remove it since we have https://github.com/qri-io/rfcs/issues/16 documenting as a record of dataset previews as a new feature

ramfox:
Would love a final review!

text/0012-CLI-commands.md
```diff
@@ -425,6 +425,736 @@ qri get [flags] <dataset_reference> <dataset_reference2> ...
...
+## qri log
+
+show log of dataset history
+
+### Synopsis
+
+
+log prints a list of changes to a dataset over time. Each entry in the log is a 
+snapshot of a dataset taken at the moment it was saved that keeps exact details 
+about how that dataset looked at at that point in time.
```
dustmop: Typo: at at

text/0012-CLI-commands.md
```diff
@@ -425,6 +425,736 @@ qri get [flags] <dataset_reference> <dataset_reference2> ...
...
 ```
 
+<a id='qri_info'></a>
+## qri info
+
+show summarized description of a dataset
+
+### Synopsis
+
+info describes datasets
```
dustmop: Can this be elaborated on, to clarify what particular info is shown? Is it exactly the same as what `list` shows (just for a single element) and if so I guess just say that?

text/0012-CLI-commands.md
b5: ooooh I like `--purge` more. `qri remove --delete me/foo` look weird. `qri remove --perge me/foo`, muy bueno.

text/0012-CLI-commands.md
dustmop: I actually didn't realize that it was possible to `use` multiple datasets at a time. Right now the text is written in the singular case everywhere: "remember a reference to a specific dataset. This dataset will..." as opposed to "remember references to a specific group of datasets. These datasets will..."

Definitely like `--add` to use an additional dataset. In addition, we could change `qri use <ref>` to `qri use <refs>` to set multiple refs at once. Still `rename` should not recognize `use`, since it's the only command where ordering has an important semantic meaning.

ramfox: Yes, totally my fault for the confusion.

Currently, `qri use` only works for `qri get`. If you look at the `qri use` documentation as it currently exists in the app, it speaks in the plural. You can `get` info on multiple datasets at once.

I was always a little suspect of having multiple references, and my bias came out when writing this, hahah. 

I assume that remembering multiple references is what we want, as that is how it is written currently, and also how `qri get` functions.

b5: yep, big fan of both multiple refs and `--add`. I vote for ratifying all this in the current RFC, and ensuring the docs refer to multiple `<refs>`.

THE RFC PROCESS IS PROVIDING CLARITY.
(plz read that in [this guy's voice](https://youtu.be/8UFIrA7hvio?t=21))

ramfox:
@dustmop @osterbit 

Need approval by another core team member before I can merge :)

Let me know if there are any other issues or if I am good to gooooooo

b5:
🎉 🌮 🎉 🌮 🎉 🌮 🎉 🌮 🎉 🌮 🎉 🌮 🎉 🌮 🎉 🌮

PR #24, opened 2018-10-16
RFC: Revise Transform Processing

b5:
This one's based on a meeting we had at Qri to deal with questions that have arisen from early user feedback. This addresses some critical design flaws in our current transform implementation.

Apologies, we're on a bit of a time crunch, I'm hoping to get this merged in time for this week's release.

### DIFF
--- a/text/0016-revise_transform_processing.md
+++ b/text/0016-revise_transform_processing.md
@@ -1,6 +1,6 @@
 - Feature Name: Revise Transform Processing
 - Start Date: 2018-11-15
-- RFC PR: <!-- (leave this empty) -->
+- RFC PR: [qri-io/rfcs#24](https://github.com/qri-io/rfcs/pull/24)
 - Issue: <!-- (leave this empty) -->
 
 # Summary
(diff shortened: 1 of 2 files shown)


stuartlynn:
This looks good to me. I think the main option you are presenting is better than the close alternative. It makes more intuitive sense to me and I think you'r correct that it means that if a transform script doesn't need to download anything, then it's easier to ignore `qri.results.download` than a passed in context object. 

Not sure about calling it `qri.results.download` though as results are intuitively what I would think of as the end of the transformation process, not an intermediary state (which it feels like the downloads are).

Just a quick tangent as well, which is not 100% relevant here, is if you are going to include an IPFS module as well for pulling in other distributed datasets? Having http is a must to interact with the existing web but it seems weird not to be able to grab a distributed dataset in the downloads phase by IPFS hash or IPFSNS, or another named Qri dataset. If you are encouraging people to use Qri as their main way of sharing data, it seems restrictive to not allow them to use those datasets in the transform of another Qri dataset. As I say probably for another RFC but wanted to mention this here.

b5:
> are going to include an IPFS module as well for pulling in other distributed datasets? Having http is a must to interact with the existing web but it seems weird not to be able to grab a distributed dataset in the downloads phase by IPFS hash or IPFSNS, or another named Qri dataset.

💯 yes. We currently have [`qri.load_dataset_head("peer/dataset")`](http://qri.io/docs/reference/starlib/#load_dataset_body) for getting other Qri datasets. When you import datasets in this way Qri automatically records the versioning info so we can later ship features to help you maintain sync across versions.

As for grabbing raw IPFS data within starlark, absolutely planned, in a few different instantiations. The first module like to rfc-then-ship is a module called `cafs` that provides an API for doing native interaction with our `cafs` interface. I have no idea if we'll ever pull this off, but by calling it `cafs` we leave the door open to something like `cafs.fetch('/ipfs/QmFoo...')` _and maybe_ `cafs.fetch('dat://4cbe2f...')`, which would fetch via if the underlying cafs instance had a DAT connection (or some way to resolve dat links). cafs can currently resolve IPNS links, so that should come for free when we ship this `cafs` module.

All this would happen in the `download` step of a transform. If other IPFS tech like IPLD stabilizes we can ship 'em as separate Qri modules for starlark.

Ideally starlark has access to both IPFS & regular internets before December 2018.

ramfox:
> 
> Not sure about calling it `qri.results.download` though as results are intuitively what I would think of as the end of the transformation process, not an intermediary state (which it feels like the downloads are).
> 

This is interesting. And I understand what you are saying, but I kind of like the fact that it seems like `qri.results` is the end of a process. It signifies to the user, without us having to explain, that you we have already run their `download` or `some_special_function` func. However, still brainstorming to think of a better word, although none has presented itself yet.

> Just a quick tangent as well, which is not 100% relevant here, is if you are going to include an IPFS module as well for pulling in other distributed datasets? Having http is a must to interact with the existing web but it seems weird not to be able to grab a distributed dataset in the downloads phase by IPFS hash or IPFSNS, or another named Qri dataset. If you are encouraging people to use Qri as their main way of sharing data, it seems restrictive to not allow them to use those datasets in the transform of another Qri dataset. As I say probably for another RFC but wanted to mention this here.

100% agree, opened an issue #26

b5:
Very good points raised about `qri.results`. I have an update on technical implementation what will affect this anyway. I started work on a [skytf branch](https://github.com/qri-io/skytf/tree/revise_tf) to confirm the plan laid out in this RFC and ran into a critical issue. An example of the issue:

```python
load("qri.sky", "qri")    # after this load call, skylark "freezes" all values in qri.sky 
                          # the "qri" value loaded should now be immutable

def download():
  return ["hello", "world"]

def transform(ds):
  qri.results.download     # <- this implies a value in the qri object is mutated :(
```

By placing return values in `qri.results`, we're violating a starlark principle. Because we have full control of how starlark behaves within the Qri environment, we _can_ cheat this, but I think it's a bad idea to go against the grain this early.

Instead, I think we should revisit the "close second" option, passing all functions a  context:

```python
def download(context):
  return ["hello", "world"]

def transform(ds, context):
  ds.set_body(context.download)
```
This is effectively moving the duty of passing state between special functions out of `qri.results` and into `context`, which is made an explicit argument. Because `context` is being passed as an argument, it's no longer considered a "frozen" value by starlark, and can be mutated.

In this example I've dropped the `results` in favor of assigning return values of special functions directly to `context.[special_function_name]`. I agree with others on "results" implying completion, which isn't nice. This feels better to me. It also means we get to take a more "planned" approach to how context works. It would be an error if a user tried to set `context.download` directly, but we could hand them an API like `context.set("val", val)` and `context.get("val")` to pass arbitrary data around.

| pros | cons |
| --- | --- |
| no longer need to `load("qri.sky", "qri")` | always have to accept a context |
| greater long term flexibility | steeper initial learning curve |
| less "magical" IMHO | we'll need to clearly document what context values will be available when |

Thoughts?

ramfox:
Womp womp.

Yes I agree this is the best choice given the constraints :)

dustmop:
How do we handle the case of downloading multiple different webpages? Return a dict with a key pointing to each?

```
def download(context):
  a = http.get("http://example.com/a")
  b = http.get("http://example.com/b")
  c = http.get("http://example.com/c")
  return {"a": a, "b": b, "c": c}
```

b5:
If's destined to be a dataset component, I think so. I've used that pattern in the past, using [surt urls](https://github.com/internetarchive/surt) as keys to create a datasets. We also use a similar pattern as a [test case](https://github.com/qri-io/dataset/blob/master/dsio/testdata/json/sitemap/body.json).

You didn't ask this, but to add some detail, a justification for passing in `context` here may also include passing around that might be destined for metadata like, `num urls requested`:
```python
def download(context):
  a = http.get("http://example.com/a")
  b = http.get("http://example.com/b")
  c = http.get("http://example.com/c")
  context.set("num_urls_requested", 3)
  return {"a": a, "b": b, "c": c}
```

This frees the return value from having to give back _all_ data, avoiding stuff like:
```python
def download():
  # ...
  return { "num_urls_requested" : 3, "data": { "a", a, "b": b, "c": c }}
```
which I think is an antipattern if we have context.

text/revise_transform_processing.md
```diff
@@ -0,0 +1,177 @@
...
+
+### How do I get state that isn't part of a dataset from one transform function to another?
+
+This extra-textual state is a question of _routing return values_. By requiring only datasets to be passed in and out, the user is forced to compromise, returning either the user input _or_ the result of network requests. In a worst-case scenario, script authors may abuse the dataset object as a kind of transitive context, stuffing download results into, say, the `meta` dataset component, and deleting it in a later. This is highly error prone.
+
+Ideally, consumers of the return value of `download` have access to whatever the author of the download function deems is relevant, and are not required to make any tradeoffs when delivering the results of network requests.
+
+
+## Proposed Change:
+I think these questions point to problems with the requirement of uniform tranform function signatures, and a predclared call order. I have concerns about how these requirements will prevent us from adapting to new challenges in the future, and don't provide enough payoff to warrent keeping them around.
```
dustmop: Typo: predclared -> predeclared

text/revise_transform_processing.md
```diff
@@ -0,0 +1,177 @@
...
+
+Let's have more support for the basic idea that functions that have different signatures _do different things_. That isn't really honored here. It's not clear from the above example that download and transform have any predeclared order or capabilities. Transform functions as implemented appear to differ in name only, and the predeclared call order runs the risk of becoming arbitrary. 
+
+To solve both of these problems, I'm proposing the following changes:
+* replace the notion of a transform function with a set of predefined _special functions_, where each function has a required signature that fits it's needs
```
dustmop: More than required signature, there's even a required *name*.

b5:
lovely. @ramfox any objections to merging this?

text/0016-revise_transform_processing.md
```diff
@@ -0,0 +1,192 @@
...
+
+`download` is an example of a _special function_. Special functions are predefined function names and signatures the Qri environment knows to look for. Unlike our current transform functions, special functions have varying signatures, which must be conveyed through documentation. If a transform script defines a special function, Qri will call it and place the return value of the function at `context.[special_function_name]` before calling the main `transform` function. The main `transform` now also has access to the result of all _special function_ calls via the passed in `context`.
+
+`context` is designed to solve the issue of moving state around during a transform. All special functions will accept a `context` argument. The Qri environment will populate `context` with necessary transfrom state such as `transform.config` and `transform.secret` values. `context` also has an API for passing arbitrary user data via `context.set("key", Value)` and `context.get("key")`.
+
```
ramfox: Does this mean that the secrets and config will be located at `context.transform.secrets`, rather then using qri.get_secrets(key)?

b5: oooh v good question. I was thinking the new API would be `context.get_config` and `context.get_secret`, both as functions that take a name, and yes, both would move off of `qri` and onto `context`.

By making them function calls it'll be easier for us to tell which config and secret values a script is making use of so we can later tell other users of the script stuff like "this transform requires the following secrets: ..."

text/0016-revise_transform_processing.md
```diff
@@ -66,41 +66,44 @@ Ideally, consumers of the return value of `download` have access to whatever the
...
 
-A side effect of these changes, transform scripts now must define a `transform` function if they wish to have any effect on a dataset. Reducing a transform to this single requirement will make for a consistant point of entry that's easier for both humans and machines to reason about.
+As a special function, `download` now has a unique signature that more closely matches it's intention. Download is intented to _download things_. It has no access to the input dataset, which is denoted by the signature. Download should do network stuff, then return the results for further processing in `transform`. Transform should modify the dataset, so it's provided a dataset param. All of this will need to be thoroughly documented, (and ideally made as a "tab completion" in the Qri frontend editor), but this change helps convey intended use.
+
+Making the return value of `download` available at `context.download` also helps clarify to the user that by the time `transform` is executing, the qri environment has already called download. This clarifies both what is doing the calling, and when it's been called.
```
dustmop: Ah, I missed this part of the change before. Though I understand the objections to using "results", I am not as excited about `context.download` as `context.{something}.download` because it feels to me like part of the hierarchy is missing; "download" isn't directly owned by "context", it's owned by something that tracks the outputs of other functions calls, which is itself owned by context. From talking with @b5, I suggested `context.from.download`. But on the other hand, this isn't a huge deal, and it would be easier to fix this in the future (introduce something like `from` and deprecate the direct `context.download` method) rather than do the opposite.

b5:
🕺 another one down!

PR #25, opened 2018-10-16
RFC: Define Dataset Creation (WIP)

b5:


text/0017-define_dataset_creation.md
```diff
@@ -0,0 +1,220 @@
...
+Simplify "new" and "save" into one algorithm that defines the high-level steps for creating a dataset, making learning Qri easier.
+
+# Motivation
+[motivation]: #motivations
+
+An early Qri user recently gave us some feedback:
+
+> when I was writing a transformation, it wasn't super clear what needed to happen in the sky file and what needed to happen in the dataset.yml. I started out with both and had to deal with a bunch of errors which stemmed from (I think) me accidentally specifying things in both places or in the wrong place. In particular I wasn't sure what runs first. For example, does the ds in the transformation know the schema specified in dataset.yml or does qri check the output of transform against the specified schema? 
+
+What I'm hearing is Qri doesn't have clear mental model of what will happen when the user runs `qri new` or  `qri save`. In some ways this makes sense, Qri is brand new. For that same reason it's important to introduce a consistent, digestable mental model as early as possible.
```
dustmop: The feedback also points to a possible ambiguity between transform and dataset.yml, since both can do similar things. This is a distinct (but admittedly related) problem from the semantics of `qri save`. I understand it's hard to answer the original question without cleaning up our mental model, but I think it's also important to address the original question directly.

text/0017-define_dataset_creation.md
```diff
@@ -0,0 +1,220 @@
...
+With Qri it's possible to just run `qri save` and follow that with `qri conenct`, and now others will be able to see and consume your dataset.
+
+Qri Publish is a far less necessary step on 
+
+-- --
+
+## Save creates a dataset snapshot
+A dataset commit is a _snapshot_ of a dataset at a given point in time.
+
+Prepare - conform input data for saving
```
dustmop: These appear on the same line in the processed markdown. Use asterisks to make them bullet pointers?

text/0017-define_dataset_creation.md
```diff
@@ -0,0 +1,220 @@
...
+Save
+
+### Steps
+* **Prepare**
+  * screen for incorrect input
+  * check for required values
+  * remove unchanged input
+  * turn a _patch_ into an _update_
+* **Run Transform**
+  * transform scripts receive a _full snapshot_
```
dustmop: "full snapshot" of what? The changes (body, meta, etc) supplied by the user, right? We should specifically call out that the previous version's body is not the dataset's body, but can be retrieved by a helper function.

text/0017-define_dataset_creation.md
b5: agreed. I've added some language that better connects the RFC content to Matt's feedback

text/0017-define_dataset_creation.md
```diff
@@ -88,36 +91,40 @@ While it is true that "qri save" is doing one less step than the git variant, gi
...
-  * to pass extra values from one step to the next, supply a tuple of return values
-  * error by calling `error()`
-  * validation is performed post-transformation
-    * use the "qri.validate" method to perform validation within a transform (currently not implemented)
-  * access versions with `qri.load_previous(ds)`
-* **Validate**
+  * transform script is passed the prepared update from the user (eg: things supplied in dataset.yaml)
+  * previous verion of input dataset (if any) is accessible via a helper function: `qri.load_previous(ds.name)`
+* **Compute Automatic Values**
+  * if no `ds.structure.schema` is supplied but body data is provided, a `ds.structure.schema` is inferred from body data. often defaulting to one of `{ "type" : "array" }` or `{ "type" : "object" }`
```
dustmop: This is excellent, very explicit about what is going on 👍

text/0017-define_dataset_creation.md
dustmop: Very good addition, but the next paragraph doesn't really start as well as it could, now that it doesn't immediately follow the "feedback" text. Maybe start it more like "But in order to properly respond to feedback such as this...."

text/0017-define_dataset_creation.md
```diff
@@ -0,0 +1,224 @@
...
+> -- @mbattifarano
+
+What I'm hearing is Qri doesn't have clear mental model of what will happen when the user runs `qri new` or  `qri save`. In some ways this makes sense, Qri is brand new. For that same reason it's important to introduce a consistent, digestable mental model as early as possible.
+
+The ambiguity around what goes in the transform and what goes in the dataset.yaml is difficult to decipher when it's not clear which is "processing first". We should make it clear that the details of dataset.yaml are processed onto the dataset _before_ the skylark file, but validation happens after.
+
+We need to clearly document _how outside data gets into Qri_ in a way that becomes natural with practice. Ideally this process has as few exceptions & "gotchas" as possible. While defining such a mental model, we've come up with changes that have 
+
```
ramfox: sentence fragment?

PR #27, opened 2018-10-18
add ci to check spelling

b5:
I'm super tired of making spelling mistakes. Let's make the computers do this. This PR adds a circleCI step that'll fail on commonly misspelled words.


b5:
it only checks "common words", which makes it skip stuff like 'DatasetRef' 😄. This branch, for example, passes without a problem, even though words like `DatasetRef` are in there.

PR #28, opened 2018-10-24
RFC: publish & update

b5:
I'd like to add these two features to Qri ASAP, as I think they help round out our sync story in a way that's both useful now and extensible later.

@dustmop ideally this RFC addresses the functionality we've been describing around `qri advance`, would love it if you could review with that use case in mind.

### DIFF
--- a/text/0018-publish-update.md
+++ b/text/0018-publish-update.md
@@ -33,16 +33,16 @@ Running `publish` with a configured registry should POST a dataset _summary_ to 
 ## Update
 Update brings your data to the latest known version of a dataset. Update comes in two flavors:
 * Updating Someone else's dataset
-* Updating A Dataset You Own
+* Updating A Dataset in your namespace
 Both types of update will be accessed via the same command like and API interfaces, they only differ based on the dataset reference in question.
 
 #### Updating a peer's dataset
 Running update on a peer dataset checks to see if the peer is online and calls `get_dataset`, if the version is different, it'll grab that dataset. If the peer in question is offline we can't check for updates. Once publish pushes to the registry we can check the registry for updates.
 
-Later on we'll want to do proactive checks for updates by submitting a list of dataset refs we'd like to grab updates for to peers and registries.
+Later on we'll want to do proactive checks for updates by submitting a list of dataset refs for which we'd like to grab updates to peers, which will let us display things like "your dataset is out of date, update?" in user interfaces.
 
-#### Updating a dataset you own
-running update on a dataset in your namespace will check if a transform script is specified, If so, it'll re-execute the transform & check for changes. If changes exist it'll create a commit.
+#### Updating a dataset in your namespace
+Running update on a dataset in your namespace will check if a transform script is specified, If so, it'll re-execute the transform & check for changes. If changes exist it'll create a commit.
 
 Manual changes to a dataset are still run with `qri save`. Calling update on a dataset with no transform script is an error. That error should tell users to run `qri save`.
 
@@ -57,12 +57,12 @@ publish and update are both steps that require an existing dataset.
 $ qri publish --help
 
 Description:
-  Qri publish makes your dataset available to others. While online peers that 
+  Qri publish makes your dataset available to others. While online, peers that 
   connect to you can only see datasets and versions that you've published. 
   publishing a dataset always makes all previous history entries available.
 
   By default publish also publishes your dataset to all configured registries,
-  to publish to only specific registries, use the --
+  to publish to only specific registries, use the --registries flag.
 
 Examples:
   # publish the latest version of a dataset
@@ -77,7 +77,7 @@ Examples:
 Flags:
   --delete -d       unpublish the specified dataset
   --no-registry -r  don't publish to the registry
-  --registries      comma separated list of registries to publish to, overrides --no-registry
+  --registries      comma separated list of registries to publish to
 ```
 
 #### API:
@@ -99,12 +99,13 @@ $ qri update --help
 
 Description:
   Update fast-forwards your dataset to the latest known version. If the dataset
-  is owned by a peer, update will ask the peer for any new versions and download
-  them.
+  is not in your namespace (i.e. dataset name doesn't start with your peername), 
+  update will ask the peer for any new versions and download them.
 
-  Calling update on a dataset you own will advance your dataset by re-running 
-  any specified transform, creating a new version of your dataset in the process.
-  If your dataset doesn't have a transform script, update does nothing.
+  Calling update on a dataset in your namespace will advance your dataset by 
+  re-running any specified transform script, creating a new version of your 
+  dataset in the process. If your dataset doesn't have a transform script, 
+  update will error.
 
 Examples:
   # get the freshest version of a dataset from a peer
@@ -131,8 +132,8 @@ POST http://localhost:2503/update/other_person/dataset
 # update your local dataset by re-running the dataset transform
 POST http://localhost:2503/update/me/dataset_with_transform
 
-# supply secrets to an update, publish on successful run
-POST http://localhost:2503/update/me/dataset_with_transform?publish=true -H secrets=keyboard,cat
+# supply secrets to an update via request headers, publish on successful run
+$curl -X "POST" -H "secrets: keyboard,cat" http://localhost:2503/update/me/dataset_with_transform?publish=true
 ```
 
 ## Library changes
--- a/text/0018-publish-update.md
+++ b/text/0018-publish-update.md
@@ -6,29 +6,37 @@
 # Summary
 [summary]: #summary
 
-Add "publish" and "update" as patterns for broadcasting when a dataset is ready for public consumption, and an update pattern for generating & syncing version histories.
+Add "publish" and "update" as patterns for broadcasting when a dataset is ready for public consumption, and an update pattern for grabbing the latest available version of a dataset.
 
 # Motivation
 [motivation]: #motivation
 
-Right now Qri is a little _too_ "public". Any time I add anything to Qri and go online, I have no tools for saying "I don't want people downloading this dataset", or "this dataset isn't ready for public consumption". This is creating "use paralysis", where I don't feel as comfortable as I should "practicing" on Qri, because any and everything I save is visible to the world. If I make a mistake, I have no clear way of telling another user "hey don't grab these updates!".
+Qri makes promises to keep user's data fresh, but we're missing explicit tools & terminology for getting latest versions. This RFC introduces two initial tools for publishing datasets and keeping versions fresh. It's an important part of our story that I'd like to ship some code against to round out our initial user experience.
+
+In regard to publishing, right now Qri is a little _too_ "public". Any time I add anything to Qri and go online, I have no tools for saying "I don't want people downloading this dataset", or "this dataset isn't ready for public consumption". This is creating "use paralysis", where I don't feel as comfortable as I should "practicing" on Qri, because any and everything I save is visible to the world. If I make a mistake, I have no clear way of telling another user "hey don't grab these updates!".
 
 Speaking of updates, I don't have any clear mechanism for re-fetching datasets that's been updated since I "added" it. We need a quick and easy way to syncronize my remote copy of a dataset with the lastest version after it's been published.
+
+Finally, much of the beauty of Qri's self-contained datasets is _not needing to understand code to make an update_. By providing users with an explicit "update" flow that re-executes transforms, that beauty is much easier to see.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-This RFC introduces two new concepts to Qri that will come with new commands and, will modify some existing behaviors. Publish and Update will work together to allow users to explicitly opt-in to sync points
-
-Because this process will be run frequently, I'd like to make the common use patterns as simple as possible, to the point that `qri save` should come with a `publish` flag that combines saving an update & publishing it into one step.
+This RFC introduces two new concepts to Qri that will come with new commands and, will modify some existing behaviors. Publish and Update will work together to allow users to explicitly specify which datasets are ready for consumption.
 
 ## Publish
-All datasets start as "unlisted", and the user must `publish` them to make visible to other users. Peers calling `list_datasets` now only see datasets you've published, at version that's been most-recently published. If I create a dataset, then publish, and save changes to that dataset, I must re-publish the new changes for others to see the update.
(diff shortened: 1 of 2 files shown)


text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+## Update
+Update brings your data to the latest known version of a dataset. Update comes in two flavors:
+* Updating Someone else's dataset
+* Updating A Dataset You Own
+Both types of update will be accessed via the same command like and API interfaces, they only differ based on the dataset reference in question.
+
+#### Updating a peer's dataset
+Running update on a peer dataset checks to see if the peer is online and calls `get_dataset`, if the version is different, it'll grab that dataset. If the peer in question is offline we can't check for updates. Once publish pushes to the registry we can check the registry for updates.
+
+Later on we'll want to do proactive checks for updates by submitting a list of dataset refs we'd like to grab updates for to peers and registries.
```
dustmop: This reads a bit awkwardly. How about "... by submitting a list of dataset refs for which we'd like to grab updates to peers..."

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+
+#### Updating a dataset you own
+running update on a dataset in your namespace will check if a transform script is specified, If so, it'll re-execute the transform & check for changes. If changes exist it'll create a commit.
```
dustmop: This makes it possible to define a transform script that breaks updates in a non-obvious way.

For example:

```
Dataset with body:
{"name": "Dustin"}
```
```
def transform(ds, context):
  prev = ds.load_previous_version()
  ds.save({"name": prev["name"] + "g"})
```

Every time this runs `update`, it'll append an additional "g", therefore it always has changes. This is true of any nearly any transform script that modifies the previous version somehow. This makes it hard to treat update like a "regularly update my sets" kind of cronjob-like task.

Perhaps instead we want `update` to only check for changes to external sources of data? Or perhaps there's a way to take advantage of our pipeline / special functions, and only follow through with changes that are generated outside of the `transform` function?

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+
+Manual changes to a dataset are still run with `qri save`. Calling update on a dataset with no transform script is an error. That error should tell users to run `qri save`.
```
dustmop: If I manually change something using `qri save` does that remove the old transform script? Or if not, and it keeps the transform script, is that script re-run each time?

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+
+publish and update are both steps that require an existing dataset.
+
+## Publish:
+#### CLI:
+```
+$ qri publish --help
+
+Description:
+  Qri publish makes your dataset available to others. While online peers that
```
dustmop: Please add a comma:
"While online, peers"

Otherwise it reads different, (while) (online peers)

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+  # publish a dataset (and all previous versions) at a specific version
+  $ qri publish me/dataset@/ipfs/QmFoo...
+
+  # unpublish a dataset
+  $ qri publish -d me/dataset
+
+Flags:
+  --delete -d       unpublish the specified dataset
+  --no-registry -r  don't publish to the registry
+  --registries      comma separated list of registries to publish to, overrides --no-registry
```
dustmop: I don't think it should override; it should be an error.

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+
+  # update your local dataset by re-running the dataset transform
+  qri update me/dataset_with_transform
+
+  # supply secrets to an update, publish on successful run
+  qri update me/dataset_with_transform -p --secrets=keyboard,cat
+
+Flags:
+  --publish, -p     publish successful updates
+  --title, -t       supply a
```
dustmop: sentence fragment

Also, why a title but not other metadata? Commit message definitely makes sense, though.

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+### API:
+```
+# get the freshest version of a dataset from a peer
+POST http://localhost:2503/update/other_person/dataset
+
+# update your local dataset by re-running the dataset transform
+POST http://localhost:2503/update/me/dataset_with_transform
+
+# supply secrets to an update, publish on successful run
+POST http://localhost:2503/update/me/dataset_with_transform?publish=true -H secrets=keyboard,cat
```
dustmop: URL encode these spaces with "+".

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+I'll work out how they work while coding.
+
+ListDatasets will have to be adjusted to read from either `repo.PublishedRefs` or `repo.Refs` depending on who's asking.
+
+### p2p:
+* `list_datasets` should now return the list of _published_ datasets.
+* `get_dataset` should only return the latest published version
+
+### repo:
+We'll need to make a new `PublishedRefs` store that is just another-but-separate `RefStore`.
```
dustmop: Alternatively, keep the same RefStore and add a flag to references that says whether they are published or not.

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+
+Speaking of updates, I don't have any clear mechanism for re-fetching datasets that's been updated since I "added" it. We need a quick and easy way to syncronize my remote copy of a dataset with the lastest version after it's been published.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This RFC introduces two new concepts to Qri that will come with new commands and, will modify some existing behaviors. Publish and Update will work together to allow users to explicitly opt-in to sync points
+
+Because this process will be run frequently, I'd like to make the common use patterns as simple as possible, to the point that `qri save` should come with a `publish` flag that combines saving an update & publishing it into one step.
```
dustmop: First time, I misread the end of this sentence as "saving and updating & publishing it..." which reminds me that "update" is still an overloaded term. An "update" is also how we refer to new snapshots applied with the `save` command, as opposed to a "patch". This could lead to confusion, or perhaps you meant the similarity of meanings to be on purpose? In that case the `qri update` command will both retrieve and create new "updates".

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+Running `publish` with a configured registry should POST a dataset _summary_ to the registry. We haven't defined dataset previews, summaries, and documents yet (a story for another RFC), but once it's done publish will both list the dataset for p2p discovery and push the dataset to the registry.
+
+## Update
+Update brings your data to the latest known version of a dataset. Update comes in two flavors:
+* Updating Someone else's dataset
+* Updating A Dataset You Own
+Both types of update will be accessed via the same command like and API interfaces, they only differ based on the dataset reference in question.
+
+#### Updating a peer's dataset
+Running update on a peer dataset checks to see if the peer is online and calls `get_dataset`, if the version is different, it'll grab that dataset. If the peer in question is offline we can't check for updates. Once publish pushes to the registry we can check the registry for updates.
```
dustmop: Can I `qri update` on a dataset that I haven't added to my own repo yet? If so does that run `qri add` under the hood first? (Note: `add` might change names in the near future, to `fork` or `clone` or something)

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+```
+$ qri publish --help
+
+Description:
+  Qri publish makes your dataset available to others. While online peers that 
+  connect to you can only see datasets and versions that you've published. 
+  publishing a dataset always makes all previous history entries available.
+
+  By default publish also publishes your dataset to all configured registries,
+  to publish to only specific registries, use the --
```
dustmop: Sentence fragment, did you mean "--registries flag"?

text/0018-publish-update.md
b5: manual changes remove a transform script by setting `transform: null`, otherwise the transform will remain. On both save and update transforms are always run. More generally, existing transforms always run.

It would be possible to hack the transform to only run in specific conditions with logic that reads a config or secret that's fed in each time, so maybe we should consider a `--no-transform` flag just to formalize it 🤷‍♂️? I'd rather wait to see if people really need the distinction.

text/0018-publish-update.md
b5: I think `qri update` on a dataset you don't have should be an error for now. updating something you don't have feels weird, but could see it coming in handy once we start to deal with doling out permissions like "dustmop can run update on this dataset". There's a whole nightmare there waiting for us 😄

text/0018-publish-update.md
b5: oooooh v good idea. that sounds nicer.

ramfox:
In office, @b5 and I had a brief convo about the word "owned" and how that is a loaded and perhaps untrue term to use for a qri dataset.  We talked about using "in your namespace"/"in a peer's namespace" instead. That is definitely more accurate, but I'm not sure its the most intuitive  phrase.

Noting via comments the places in the rfc that would  need to be changed from "owned" to a different term.

text/0018-publish-update.md
ramfox: "Updating A Dataset In Your Namespace"?

text/0018-publish-update.md
ramfox: "Updating a dataset in your namespace"

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+## Update:
+#### CLI:
+```
+$ qri update --help
+
+Description:
+  Update fast-forwards your dataset to the latest known version. If the dataset
+  is owned by a peer, update will ask the peer for any new versions and download
```
ramfox: "If the dataset is not in your namespace, update will ask the peer"

"If the dataset is in a peer's namespace, update will ask that peer"

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+  them.
+
+  Calling update on a dataset you own will advance your dataset by re-running
```
ramfox: "Calling update on a dataset in your namespace will..."

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+
+Whenever I choose, I may "unpublish" a dataset, which will remove it entirely from the list returned by `list_datasets`. It is true that others may still have a copy of my dataset. At present there's not much I can do about it.
+
+#### Publish will integrate with Registries
+Running `publish` with a configured registry should POST a dataset _summary_ to the registry. We haven't defined dataset previews, summaries, and documents yet (a story for another RFC), but once it's done publish will both list the dataset for p2p discovery and push the dataset to the registry.
+
+## Update
+Update brings your data to the latest known version of a dataset. Update comes in two flavors:
+* Updating Someone else's dataset
+* Updating A Dataset You Own
```
ramfox: "Updating a Dataset In Your Namespace"

text/0018-publish-update.md
```diff
@@ -0,0 +1,180 @@
...
+Both types of update will be accessed via the same command like and API interfaces, they only differ based on the dataset reference in question.
+
+#### Updating a peer's dataset
+Running update on a peer dataset checks to see if the peer is online and calls `get_dataset`, if the version is different, it'll grab that dataset. If the peer in question is offline we can't check for updates. Once publish pushes to the registry we can check the registry for updates.
+
+Later on we'll want to do proactive checks for updates by submitting a list of dataset refs we'd like to grab updates for to peers and registries.
+
+#### Updating a dataset you own
```
ramfox: "Updating a dataset in your namespace"

text/0018-publish-update.md
b5: You're right, the term is overloaded, and I didn't mean for previous RFC's version of `update` to collide with this one. 

I now think the former RFC's use of the phrase "update" was a mistake, and instead should be called something that refers to dataset _completeness_, instead of referring to _temporality_. I'm working on an RFC that'll hopefully clarify terms around the "dataset completness" question. I'll cite that poor use of the word `update` as a need for such a document.

b5:
Oookay, I've hit a roadbloack that came from starting to code this RFC up.

@dustmop brought up the idea of adding a flag to indicate if a dataset version is published or not, removing the need for two lists. We can't use this approach, and the reqirement for a separate list highlights an implicit assumption about the way Dataset References work that we need to address:

We can't use this approach because _refstore assumes only **one** alias of a dataset name will be in the store at at time_. We inherited this thinking from git, in which a _ref_ is a "pointer to tip". In Qri an _alias_ is a pointer to the latest version (`peername/datset_name` is an "alias"). Having multiple references with different hashes but the same alias in one Refstore will lead to undefined behaviour.

For publiched to work properly, we need to be able to store multiple references to the same alias, because "published" and "tip" need to deviate in hash, but not alias:

```
versions:    a ---- b ---- c ---- d
                    |             |
published:          *             |
tip:                              *

published ref:  b5/dataset@QmProfileID/ipfs/b
tip ref:        b5/dataset@QmProfileID/ipfs/d
```

After working through this question, and considering the above diagram, it's clear that **this RFC as framed is sneaking the notion of a "branch" into Qri, under the guise of calling them "published"**. In git a branch is a one of potentially many version histories in a repo. Here we're going from one version history to two: published and tip. This is a slippery slope.

For now I'm going to adjust the RFC to state that datasets are either published or not, going with @dustmop's suggestion. There will be no way to have some part of a version history published, and another part unpublished. It's less functional, but it's a start that I think helps solve the main pain point of wanting a dataset to default to being "unlisted".

Feeling pretty pumped about the RFC process rn for pointing this out ahead of time.

b5:
alright, re-framed. I think this is ready for review

text/0018-publish-update.md
```diff
@@ -6,38 +6,46 @@
...
+
 ## Update
 Update brings your data to the latest known version of a dataset. Update comes in two flavors:
 * Updating Someone else's dataset
 * Updating A Dataset in your namespace
 Both types of update will be accessed via the same command like and API interfaces, they only differ based on the dataset reference in question.
 
 #### Updating a peer's dataset
-Running update on a peer dataset checks to see if the peer is online and calls `get_dataset`, if the version is different, it'll grab that dataset. If the peer in question is offline we can't check for updates. Once publish pushes to the registry we can check the registry for updates.
+Running update on a peer dataset checks to see if the peer is online and calls `get_dataset`, if the version is different, it'll grab that dataset. If the peer in question is offline we can't check for updates. Once `publish` also writes to registries, we can check the registry for updates to help aleviate the avilability problem.
```
dustmop: avilability -> availability

ramfox:
LGTM

PR #30, opened 2018-11-06
RFC: Distinguish Manual & Scripted Tranforms

### LINKED ISSUE #29: RFC request: transform & body semantics
Larger write-up is being put aside from now, but some notes from a large discussion today around transformation semantics:

* do transforms always run?
* how do transforms and body's provided at the same time interact?
* our current model treats the input dataset to a transform as the user provided body, is this best?
* when is it ok to drop data, such as bodies provided to transformation scripts that get dropped without being recorded everywhere. for things like `reduce` we probably want this

General outcome:

* yes, transforms always run
* we should consider things like the "--body" flag to be a "user-transformation", and running save with both "--transform" and "--body" is trying to run two transformations at once
* instead, the body should always override the transformations output, with a warning or error about changing both
* transform might need another concept, like "user-input.json", that take the role that body passed into transform currently fills

b5:
While I'm writing this RFC, this one is wholly motivated by internal discussions initiated by @dustmop, the solution to this comes from a seed idea of thinking as manual manipulations as "human transforms" planted by @ramfox, and user feedback from @mbattifarano. Thanks to all for speaking up and helping me understand how we can make Qri work better.

This RFC seems like it's complicating things. I'd argue it's making stuff that used to be scary-bad-magic explicit.

closes #29

### DIFF
--- a/text/0020-distingush_manual_vs_scripted_transforms.md
+++ b/text/0020-distingush_manual_vs_scripted_transforms.md
@@ -1,64 +1,86 @@
 - Feature Name: Distinguish Manual & Scripted Tranforms
-- Start Date: <!-- (fill me in with today's date, YYYY-MM-DD) -->
-- RFC PR: <!-- (leave this empty) -->
+- Start Date: 2018-11-05
+- RFC PR: https://github.com/qri-io/rfcs/pull/30
 - Issue: <!-- (leave this empty) -->
 
 # Summary
 [summary]: #summary
 
-Distinguish between manual and scripted transforms to ensure reproducible dataset histories, reduce confusion and overlapping responsibilities.
+Enforce a mutually-exclusive distinction between manual and scripted transforms to ensure reproducible dataset histories, reduce confusion and remove overlapping responsibilities.
 
 # Motivation
 [motivation]: #motivation
 
+Qri has a goal of being as simple as possible, simple can be measured in two ways: 
+
+1. How _concise_ something is: a thing having few steps, requiring fewer instructions
+2. How _clear_ something is: knowing what will happen when you interact with that thing.
+
+Here's some feedback from @mbattifarano an early Qri user. We've cited this in [another other rfc](https://github.com/qri-io/rfcs/blob/master/text/0017-define_dataset_creation.md), but I don't think we're done mining it for cues to improve:
 
 > The role of dataset.yml was a bit confusing to me because it seems to have a few responsibilities some of which are not unique to it. Two concrete examples: first, when I was writing the transformation to create me/fare_per_pax, it wasn't super clear what needed to happen in the sky file and what needed to happen in the dataset.yml. I started out with both and had to deal with a bunch of errors which stemmed from (I think) me accidentally specifying things in both places or in the wrong place. In particular I wasn't sure what runs first. For example, does the ds in the transformation know the schema specified in dataset.yml or does qri check the output of transform against the specified schema?
 
+@dustmop pointed directly at the issue this feedback raised [while we were working on that rfc](https://github.com/qri-io/rfcs/pull/25#discussion_r225723031):
 
-One of many not-so-nice surprises about the way Qri currently works comes from the passed-in dataset in a starlark transform:
+> The feedback also points to a possible ambiguity between transform and dataset.yml, since both can do similar things. This is a distinct (but admittedly related) problem from the semantics of `qri save`. I understand it's hard to answer the original question without cleaning up our mental model, but I think it's also important to address the original question directly.
+
+Both are right. Words like "confusion" and "ambiguity" mean lack of clarity. Qri is emphasizing brevity too much, and needs to be more clear, even if it means adding a step or two. To do this properly, each added step should reinforce the mental model of _applying changes to a dataset_.
+
+There's also  a technical motivation for making this change as well What's worse, this "too many things" is costing us reproducibility. Because multiple things can act on a dataset to get form one snapshot to another, we can't accuractly reproduce histories by re-playing transformations. 
+
+## Proposal distinguish between "Manual" and "Scripted" transforms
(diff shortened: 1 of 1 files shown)


text/0020-distingush_manual_vs_scripted_transforms.md
```diff
@@ -0,0 +1,312 @@
...
+updated dataset b5/ca_prime_minsters
+```
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Human and Machine Transforms that affect the same field is now an error
+
+While this RFC places the definition at the field level, our initial implemenation will operate at the component level for simplicity.
```
dustmop: implemenation -> implementation

text/0020-distingush_manual_vs_scripted_transforms.md
```diff
@@ -0,0 +1,312 @@
...
+
+@dustmop pointed directly at the issue this feedback raised [while we were working on that rfc](https://github.com/qri-io/rfcs/pull/25#discussion_r225723031):
+
+> The feedback also points to a possible ambiguity between transform and dataset.yml, since both can do similar things. This is a distinct (but admittedly related) problem from the semantics of `qri save`. I understand it's hard to answer the original question without cleaning up our mental model, but I think it's also important to address the original question directly.
+
+Both are right. Words like "confusion" and "ambiguity" mean lack of clarity. Qri is emphasizing brevity too much, and needs to be more clear, even if it means adding a step or two. To do this properly, each added step should reinforce the mental model of _applying changes to a dataset_.
+
+There's also  a technical motivation for making this change as well What's worse, this "too many things" is costing us reproducibility. Because multiple things can act on a dataset to get form one snapshot to another, we can't accuractly reproduce histories by re-playing transformations.
```
dustmop: missing punctuation after "as well"

text/0020-distingush_manual_vs_scripted_transforms.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Proposal distinguish between "Manual" and "Scripted" transforms
+To me, the root of the problem here is we're being concise as the cost of clarity. it's clear that Qri is doing too many things in a single step, and allowing multiple things to affect the outcome of a `qri save`.  To solve this, we'll enforce some new rules and ratchet up some definitions. I want to start by adding formality to the the term "transform":
+
+### A "transform" is a forward transition from one dataset snapshot to another.
+* transforms must mutate one or more non-computed fields of a dataset
+* only one type of transform can be applied to any field per transform
+* transforms can use one or more types of mutations to determine the next snapshot
+
+A transform is the opposite of a history, moving forwards in snapshots instead of backwards. A history is _reproducible_ when you can start at the first snapshot and re-execute each mutation described in the next snapshot. By enforcing the mutually-exclusive mutations, each snapshot is a deterministic record of both state and _how to arrive at that state_ from the previous snapshot.
```
dustmop: I think it's important to add the assumption that external resources are unchanging. For example, we must assume that fetching some file over http will have the same result in the future. Of course, sometimes this assumption will fail, but that's outside of our control and there's nothing we can do about it. But as long as we either have datasets that don't rely on external resources and/or the assumption of unchanging externalities holds, we guarantee reproducibility.

text/0020-distingush_manual_vs_scripted_transforms.md
```diff
@@ -0,0 +1,312 @@
...
+
+```python
+def transform(ds, ctx):
+  ds.set_meta("title", "rational number series")
+  # use the range function to automate setting the dataset body
+  ds.set_body(range(1,11))
+```
+
+### Transform types are mutually exclusive
+Both types of transform are acting on the same components and fields of a dataset (`meta` and `body` components, a `title` field and body `rows`). To ensure reproducibility, we need a new rule: **only one type of transform can mutate a field between two snapshots**.
```
dustmop: I would prefer to phrase this as "each type of transform must only mutate the dataset in a way that is composable with all other transforms". This is more future proof, as it eventually lets us weaken the current reality.

text/0020-distingush_manual_vs_scripted_transforms.md
```diff
@@ -0,0 +1,312 @@
...
+
+
+## `save` is for humans, `update` is for machines
+Save is now the more powerful of the two commands. 
+
+By default save will _not_ recall and re-run transforms, but will run provided transforms. running `qri save` without providing a tra
+
+Local updates will need some refinement. Running `qri update me/dataset` will now only succeed if a 
+Update gets a new flag: `--recall-tf`
+
```
dustmop: Please add an answer for a question raised in the earlier feedback: when does the body get validated against the schema? I believe it's only after the transforms are all applied, but that should be explicitly stated here, since it gets brought up at the start of the RFC.

text/0020-distingush_manual_vs_scripted_transforms.md
```diff
@@ -0,0 +1,312 @@
...
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The alternative is what we're currently doing, which is a sort of simultaneous three-way merge between the previous commit, user input, and script results. A more realistic alternative would be further refinements on this model, which we'll hopefully figure out through use.
+
+# Prior art
+[prior-art]: #prior-art
+
+[Git Hooks](https://githooks.com/) were mentioned as an example of prior art. There isn't much to go on here.
+There is no analog for a transform script in git. Git hooks are the closest concept, but only allow the user to receive notifications and interrupt the commit process, not affect the commit itself
+
```
dustmop: Also, macros in spreadsheet programs like Excel, Google Sheets, and OpenOffice have some analogue to what we're doing.

text/0020-distingush_manual_vs_scripted_transforms.md
b5: lovely. I think both can/must be true. Will add.

b5:
slammin

PR #34, opened 2019-01-23
rfc, export: Export behavior explained in more detail

dustmop:


text/0018-export_behavior.md
```diff
@@ -0,0 +1,149 @@
+- Feature Name: Export Behavior
+- Start Date: <!-- (fill me in with today's date, YYYY-MM-DD) -->
+- RFC PR: <!-- (leave this empty) -->
+- Issue: <!-- (leave this empty) -->
+
+# Summary
+[summary]: #summary
+
+Export functionality exists in qri, but is lacking many options and pieces of functionality.
```
b5: Ooook @dustmop & I spent a bunch of time hammering this out IRL and came up with some major revisions, best expressed as refinements to the semantics of our commands:

#### `qri get`
show a component of a dataset. Use `--format` to specify the output format. Giving format conversion to `qri get` answer this use case:
> * Getting the body as json or csv and writing to a file, the opposite of `save`.
This will now be:
```
qri get body --format csv me/dataset
```
The upside is this can also apply to other components `qri get meta --format yaml me/dataset` will be useful too. the limitation of `qri get` is you can't use it to get the entire dataset. Get without a selector *`qri get me/dataset`)

#### `qri export`
Create an external copy of a dataset. Fidelity of the exported file depends on output format. _High fidelity_ exports are immutable copies of raw blocks. High fidelity exports can be imported to qri with `qri add`, which doesn't create a commit, instead injects data directly into a qri repository. This supports the use case:
> * Compressing into a zip file for storage / transportation / archive.
> * Exporting with the intent to import later on.

High fidelity exports are directories that contain a`manifest.cbor` and sharded directory of blocks by default. We discussed the idea of defaulting to  using a custom extension `.qri`. I'm not super down with that ATM.

A new flag `--takeout` is a high fidelity .zip archive that includes the full dataset history of json files named by timestamp, and a directory of blocks for the full history. (the closest thing to `git clone`. maybe this flag could be named `--clone`?)

The default `qri export me/dataset` is equivalent to `qri export --format=raw --zip me/dataset`, yielding a high fidelity export of the latest version of a dataset.

_low fidelity_ exports put qri datasets into common formats (JSON, XLSX, CBOR, directory of files). We call these formats "low fidelity" not because they're worse, but because their default behavior is not a complete representation of a dataset. Low fidelity exports serve the important use case of external modification:
> Exporting an entire dataset with metadata and body to analyze in a script. A single file "json" or "yaml", for example.
Low fidelity exports can be used to modify a file in an external host program and ingested into qri with `qri save`. This will create a new commit and places all external modifications in the "manual transform" family of version transitions. Low fidelity exports typically _don't_ include transform scripts, as they really only make sense in the qri environment. If this changes we can revisit.

To get a low fidelity export, simply supply the desired format with the `--format` flag, any format other than `raw` is low fidelity. `qri export --format=json` will create a flat dataset with a json body and the `structure` conformed to match.

We also discussed giving `qri export` the same `--revisions` flag that delete now uses to specify subsections of history, seems rational to me. Using `qri export --format=CBOR --revisions=3 me/dataset` would yield three CBOR files, named according to the file naming convention outlined in this RFC. 

#### `qri add`
add exiting dataset to your local repository. The only thing that changes here is `qri add` will only accept high fidelity exports.`qri add --file=me_dataset.zip` will work if `me_dataset.zip` is the output of `qri export me/dataset`

b5: hrm after sleeping on it maybe "high fidelity" and "low fidelity" should switched out for "lossless" and "lossy"

text/0018-export_behavior.md
```diff
@@ -0,0 +1,149 @@
...
+
+Here are some of the use cases that pop into mind:
+
+* Getting the body as json or csv and writing to a file, the opposite of `save`.
+* Exporting an entire dataset with metadata and body to analyze in a script. A single file "json" or "yaml", for example.
+* Compressing into a zip file for storage / transportation / archive.
+* Exporting with the intent to import later on.
+* Converting to a binary format (like xlsx) for use in another program.
+
+Considering these possible use cases, the one that most seems to fit the "vanilla" behavior of `export`, what happens when no options are specified, is to export a dataset with metadata and body as a single file.
```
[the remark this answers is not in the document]
dustmop: I'm good with components as well.

text/0018-export_behavior.md
```diff
@@ -0,0 +1,145 @@
...
+qri export --format json me/my_dataset
+```
+
+```
+qri export --format xlsx me/my_dataset
+```
+
+A file created by these commands still contains as much data about the original dataset, but reimporting it is not guaranteed to end up with the exact same result as the dataset it was exported from.
+
+For the case of "json", the result will be a single file, with the body stored inline in the document.
```
b5: Can we get a quick addition here clarifying that the only way to re-import a foreign export is to create a new version of a dataset using `qri save`?

text/0018-export_behavior.md
```diff
@@ -0,0 +1,145 @@
...
+* Deciding to stop using qri, and wanting to get all of your data out in a high-fidelity form, with both machine and human readable formats.
+
+These should all be doable using `export`.
+
+Some non-goals:
+
+* Retrieving only a single component of a dataset, like the title.
+* Converting a dataset's body to a different format than it is being stored as (csv to json or cbor).
+
+These types of operations should be added to `get` instead, which already has similar functionality.
```
b5: I feel like we should should note that what binds all these use cases together is an effort to have a complete  dataset. `get` is intended for working with components of datasets. I feel like it should be a non-goal of export to support flags for removing parts of datasets from export, and should instead focus on having that functionality in `get`

PR #36, opened 2019-02-21
rfc, remotes: Remotes are like a do-it-yourself Registry

dustmop:


text/0022-remotes.md
```diff
@@ -0,0 +1,98 @@
...
+* Over time, move the rest of the functionality from the Registry into the backend remote.
+* Replace the live Registry with a standard remote.
+* Delete old Registry code.
+* New semi-centralized features get added to the Remote feature set.
+
+Some specific notes as to how this is presented. First, the concept of the Registry is a good way to explain Qri to users: it exists as a way to keep data online even if your own node is not. This shouldn't change as we move to Remotes; instead of deleting the _concept_ of a Registry, we're generalizing it by introduced a Remote as something you can run yourself if you're an advanced user with a special use-case. Secondly, since a Remote is considered an expert-level concept, it's okay for its UX to be slightly more advanced than it may be with other features. The easy thing to use should remain to be the Qri run public Registry. This implies, for example, that setting up a client to use a remote should be more like `git config set remote <url>` instead of `git remote add <name> <url>`, with `remote` as a top-level command.
+
```
b5: I think we should add an explicit section here

### Access Control / Remote Rulesets
Currently, we don't have a robust system for declaring rules about what datasets a remote should accept. Defining these rules will require additional RFCs to map the problem space before dev work can begin.

Until robust ruleset declaration exists, remotes will operate as an opt-in experiment for Qri users, with very limited means of configuring what data to accept. Users operating a remote should be prepared to manually monitor the remote they're operating for data added there.

text/0022-remotes.md
```diff
@@ -0,0 +1,98 @@
...
+Currently, the Registry exists to solve multiple problems centered around a distributed system's lack of centralization:
+
+* User identity and registration
+* Search
+* Availability
+
+The Registry acts in a "semi-centralized" way: it provides these types of features, but is not strictly required to use Qri by itself. However, there is currently no way for an average user to run something like the Registry on their own.
+
+Currently functionality that exists only in the Registry now includes:
```
b5: I think everything from here to the end of the "Guide level explanation" section should move into the "Reference Level" section.

text/0022-remotes.md
```diff
@@ -0,0 +1,98 @@
...
+* Backend - a Qri program that is not the Electron frontend
+
+Currently, the Registry exists to solve multiple problems centered around a distributed system's lack of centralization:
+
+* User identity and registration
+* Search
+* Availability
+
+The Registry acts in a "semi-centralized" way: it provides these types of features, but is not strictly required to use Qri by itself. However, there is currently no way for an average user to run something like the Registry on their own.
+
```
b5: Let's add a section here on how the proposed workflow would look:

### Propsed workflow
Assuming a new Qri remote is up and running, users who wish to push to a remote would have to do three things:
1. find the URL of a running remote, (let's pretend a remote is running at `https://example.com`)
2. run `qri config add remotes.eg https://example.com`. This would create a new remote named `eg`
3. publish a dataset using `qri publish --remote eg` 

Runnig the default `qri publish` would still push to the default registry. Users should be able to publish to multiple remotes with many `remote` flags, or by separating remote names with commas.

Users can pick whatever name they would like for remotes. Remotes will initially need to be set up with DNS records we'll describe in detail later. Setting up a remote will be an advanced thing for a while.

b5:
@camhunt, after our conversation we Qri folk spent some time thinking through the "can we run our own registry" question, and we came up with this remotes RFC. This starts with adding the capacity to "push" datasets to other places. We'd like to use this as a foundation for building some of the other features surrounding trusted peer coordination we discussed. Any feedback would be appreciated!

b5:
Data together friends (@frijol, @meiqimichelle, @titaniumbones, @lightandluck), if we get this into Qri this would let us run the Compute Canada Data Together node as a remote, which would mean `qri publish --remote data-together` would push to that infrastructure without needing to go through a server Qri runs, helping prove our "no single point of failure" story. 

I'm cc'ing y'all in case you're interested in tracking progress, or would like to steer the design of this feature. I'm delighted to hear people want the capacity to aggregate their own data 😄.

text/0022-remotes.md
```diff
@@ -0,0 +1,98 @@
...
+# Summary
+[summary]: #summary
+
+Remotes act as a way for any user of Qri to setup their own server that keeps datasets alive, providing availability and ownership over data within a set of nodes that they control.
+
+# Motivation
+[motivation]: #motivation
+
+Multiple users have requested a way to keep datasets alive and available within their own network. Currently, we have a public Registry that serves a similar purpose, but it acts too centralized and is also not designed to be duplicated and deployed by existing users. Although IPFS may keep data blocks alive due to its distributed nature, there's no guarantee to keep data around forever unless it is pinned, and the pinning node remains online. Remotes solve this problem by giving control to users, letting them own their data. Having it work as a do-it-yourself service goes a long way torwards turning Qri into a descentralized (as well as distributed) service.
```
Frijol: typo: descentralized > decentralized

text/0022-remotes.md
```diff
@@ -0,0 +1,98 @@
...
+Relatedly, the concept of the Registry, as it is now, should be deprecated and eventually removed. What is acting as the Registry now should be moved over to a normal Qri backend node that is acting as a remote.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Some definitions used in this document:
+
+* Registry - the Qri-run service for pinning, searching, etc
+* Remote - the future functionality included in the main Qri binary
+* Client - a Qri command-line program that get commands from stdin
```
Frijol: typo: get > gets

text/0022-remotes.md
```diff
@@ -0,0 +1,98 @@
...
+* Backend - a Qri program that is not the Electron frontend
```
Frijol: not sure what this means

text/0022-remotes.md
```diff
@@ -0,0 +1,98 @@
...
+
+* `POST /dataset` to upload a dataset's head
+* `/search` to search over known datasets (online or not)
+* `/datasets` to list all known datasets (online or not)
+* `/profile` user information
+* `/reputation` user trustability
+* `/pins` pinning a dataset
+* `/dsync` data block uploading
+
+Long-term we want to obliviate the need for these to be in a separate executable.
```
Frijol: typo obliviate > obviate

text/0022-remotes.md
```diff
@@ -0,0 +1,98 @@
...
+* DAGSync occurs (or not, if the dataset body is rejected)
+* Dataset is pinned
+* Remote tells the client about the result
+
+In the future, advanced users may want to configure a Remote with full granularity how an acceptance ruleset works. Perhaps they could even use `starlark` as a way to decide this.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Centralization is bad, which is why we need to be careful to make Remotes remain an optional feature that users are not forced into.
```
Frijol: Don't remotes create a federated model rather than a centralized model, though? Is Qri strongly opinionated that only p2p is a good idea here?

dustmop:
Thanks for the feedback @b5 and @Frijol! I updated the text on Friday that incorporates your comments and hopefully clarifies some of what's going on.

> At that level, my question is: are these registry nodes (here called remotes) functionally different from the Qri Registry?

The idea is that a "remote" is something that anyone can run to keep their data alive and provide features like search and user identity within a limited scope, while the "registry" is only run by Qri, and acts as the default hub that also provides such features globally. Hmmm, perhaps I can find somewhere to explicitly call out this difference in the rfc itself...

titaniumbones:
Hey, thisi s super-interesting and I am following along.  Too braindead to offer much in the way of substantive response but just flagging my interest here.

camhunt:
@b5 this initial approach for remotes helps mitigate a near-term challenge we are facing, and we'd start using it immediately.

Has there been any thought/discussion in leveraging gossipsub so that any node that subscribes to a channel will automatically replicate any data set advertised? 

This doesn't solve the bad actor problem (which we could solve - crudely - by a whitelist of nodes which any given node would auto-replicate new data sets), but it would be a more decentralized/distributed solution. When combined with IPFS public proxies, it would even mean that nodes that can't directly connect to each other (e.g., due to NAT issues) could still replicate data sets around (per gossipsub topic, assuming the public proxy is also gossiping).

b5:
@camhunt 

> this initial approach for remotes helps mitigate a near-term challenge we are facing, and we'd start using it immediately.

Great to hear. We're very much thinking of this as a first step, I'll consider you a yay vote on ratifying this RFC.

Regarding subscribing to datasets: I've started a discussion over on #37 that we can start turning into an RFC in the coming days. feedback welcome!

text/0022-remotes.md
b5: Agreed, this remotes RFC creates something that looks a lot like a federated model, and is also less centralized than the way Qri currently works. This RFC is moving more privileged capabilities onto peers.

The only thing that beats a "pure decentralized" approach for us is the requirement that it work consistently. So no, we're not "strong p2p" here, but will head in that direction whenever we can 😄. We'll add some clarifying details here.

b5:
Ok, just pushed a few spelling tweaks & a moved paragraph. I'm a 👍 on merging this. You in @ramfox & @dustmop?

text/0022-remotes.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Proposed workflow
+
+Assuming a new Qri remote is up and running, users who wish to push to a remote would have to do three things:
+
+1. Find the URL of a running remote, (let's pretend a remote is running at `https://example.com`)
+2. Run `qri config add remotes.eg https://example.com`. This would create a new remote named `eg`
+3. Publish a dataset using `qri publish --remote eg`
+
+Runnig the default `qri publish` would still push to the default registry. Users should be able to publish to multiple remotes with many `remote` flags, or by separating remote names with commas.
```
ramfox: runnig => running

b5:
nice. ok that's a majority approval from core team. LGTM!

PR #38, opened 2019-03-16
rfc: load_dataset

b5:
I've had some time to live with starlark now, and a number of use cases that have cropped up where depending on a dataset with `qri.load_dataset` has been problematic. I initially thought this was a permissions problem, describing the issue as required "permission escalation" to allow access to datasets during the `download` special function.

While working through this RFC, I've come to believe the problem is _the way we declare dataset dependencies_. This RFC aims to clear that up while pushing the problem of access control down the road, but still make some progress on _how_ access control will integrate with this RFC.

### DIFF
--- a/text/0023-starlark_load_dataset.md
+++ b/text/0023-starlark_load_dataset.md
@@ -1,26 +1,26 @@
 - Feature Name: starlark load_dataset
 - Start Date: 2018-03-15
-- RFC PR: <!-- (leave this empty) -->
+- RFC PR: [#38](https://github.com/qri-io/rfcs/pull/38)
 - Issue: <!-- (leave this empty) -->
 
 # Summary
 [summary]: #summary
 
-Introduce a global function `load_dataset` to starlark scripts both loads the dataset and declares it as a dependency. Deprecate all `qri.load_dataset` functions, making `load_dataset` the only way to depend on an external dataset.
+Introduce a global function `load_dataset` to starlark scripts that both loads the dataset and declares it as a dependency. Deprecate all `qri.load_dataset` functions, making `load_dataset` the only way to depend on an external dataset.
 
-This RFC assumes all datasets are public. In the future `load_dataset` will be expanded, accepting optional arguments to scope the kinds of access a script is requesting.
+This RFC assumes all datasets are public, and that names refer to a dataset the script author can resolve. In the future `load_dataset` will be expanded, accepting optional arguments to scope the kinds of access a script is requesting.
 
 # Motivation
 [motivation]: #motivation
 
-Qri's goal is to build an environment where datasets are as a first class citizen. Qri currently monitors any call to `qri.load_dataset` and records the loaded dataset as a dependency. This results in a complete dependncy graph.
+Qri's goal is to build an environment where datasets are as a first class citizen. Qri currently monitors any call to `qri.load_dataset` and records the loaded dataset as a dependency. This results in a complete dependency graph.
 
 There are (at least) three problems with this approach: 
 
 * **scattered requirements**
-  Dependencies are not explicit in code. The only way to know what datasets a script depends on is to read the entire script. We've learned this lesson many times with software dependencies, and know to explicitly require dependency-forming statements to be at the top of a code file. There's no reason this shouldn't apply to depending on datasets.
+  Dependencies are not explicit in code. The only way to know what datasets a script depends on is to read the entire script. We've learned this lesson many times with software dependencies, and know to explicitly require dependency-forming statements at the top of a code file. There's no reason this shouldn't apply to depending on datasets.
 * **inexpressive usage**
-  `qri.load_dataset` functions are is not expressive enough to capture all the uses for a dataset, building up patternas around loading pieces of a dataset inside special functions like `transform` is missing the point of thinking of datasets as _documents_.
+  `qri.load_dataset` functions are not expressive enough to capture all the uses for a dataset, building up patterns around loading pieces of a dataset inside special functions like `transform` is missing the point of thinking of datasets as _documents_.
 * **expanding adds to `qri` module, not dataset documents**
   Currently the `qri` module returns plain values from methods like `load_dataset_body`. If we wanted to add a function that, say, only selected specific parts of a dataset body, that function would be added to the `qri` module. This issue will only be exacerbated when we try to overlay access control onto datasets. It makes more sense to define all ways a dataset will be used at the point of import, and have Qri return a `dataset` document that is scoped to that use.
 
@@ -74,7 +74,9 @@ residents = load_dataset("b5/city_residents", anonymized=True)
 residents = load_dataset("b5/city_residents", forward_permissions=True)
 ```
 
-These are suggestions only. We need lots of time to think through how these options will work, this rfc only adds the assumption that those statements will be expressed as options on `load_dataset`. We can build on the already-present assumption that loaded datasets are read-only, making all options pertain to read access.
+These options are not giving additional permissions to scripts, as that would go against the no-need-to-audit goal, they are instead self-restricting what the loaded dataset may be used for. They exist for the benefit of the transform author, not for the security of the transform consumer.
+
+These are suggestions only. We need lots of time to think through how these options could work, this rfc only adds the assumption that those statements will be expressed as options on `load_dataset`. We can build on the already-present assumption that loaded datasets are read-only, making all options pertain to read access.
 
 We _do_ need to declare what the default `load_dataset` expression means in relation to permissions:
 
@@ -119,13 +121,16 @@ def transform(ds, ctx):
   ds.set_body([ds.get_meta('title') for ds in datasets])
 ```
 
-Statements that use scropts to compose import names, however, can and should work:
+This example is making use of a _privileged resource_ (the local qri node), which should be denied by default. For this assumption to hold, privileged resources must _only_ be available within special functions, an assumption we already enforce.
+
+Statements that use logic to compose import strings should also be denied:
 
 ```python
+# error: load_dataset must be called with a string value
 pop_years = [load_dataset("b5/population_%s" % year) for year in range(2011,2017)]
 ```
 
-The former example is making use of a _priveledged resource_ (the local qri node), while the latter is a convenience function for creating legal string names. For this assumption to hold, priveledged resources must _only_ be available within special functions, an assumption we already enforce.
+This convenience function for creating legal string names breaks static analysis tools, making it impossible to determine a script's dependencies without executing code.
 
 ### Name Resolution
 When the starlark execution environment encounters a `load_dataset` statement, it has to _resolve_ the name, connecting the string reference to a dataset object. This may or may not require network access, depending on if the user currently has all of the specified versions locally. All naming duties required by `load_dataset` will be offloaded to Qri's name resolution system, but a few interactions merit calling out here:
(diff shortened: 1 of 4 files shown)


text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,249 @@
...
+
+# apply differential privacy to all statements that read from the body
+residents = load_dataset("b5/city_residents", anonymized=True)
+
+# require that the permissions of the dataset that results from this transform
+# match the permissions of "b5/city_residents"
+residents = load_dataset("b5/city_residents", forward_permissions=True)
+```
+
+These are suggestions only. We need lots of time to think through how these options will work, this rfc only adds the assumption that those statements will be expressed as options on `load_dataset`. We can build on the already-present assumption that loaded datasets are read-only, making all options pertain to read access.
```
dustmop: Important to note: these options are not giving additional permissions to scripts, as that would go against the no-need-to-audit goal, they are instead self-restricting what the loaded dataset may be used for. They exist for the benefit of the transform author, not for the security of the transform consumer.

text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,249 @@
...
+### `load_dataset` must be top level scope
+`load_dataset` statements will only be allowed within the primary scope of a transform script. Phrased another way: `load_dataset` _cannot_ be called within a function. This limits the scope of what is possible when composing dynamic import statements. This **won't** work:
+
+```python
+load("qri.star", "qri")
+
+# create a dataset of all dataset titles
+def transform(ds, ctx):
+  # error: load_dataset must be called in top level scope
+  datasets = [load_dataset(name) for name in qri.list_datasets()]
```
dustmop: Great!

text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,249 @@
...
+  ds.set_body([ds.get_meta('title') for ds in datasets])
+```
+
+Statements that use scropts to compose import names, however, can and should work:
```
dustmop: Spelling: scropts -> scripts

text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,249 @@
...
+
+```python
+pop_years = [load_dataset("b5/population_%s" % year) for year in range(2011,2017)]
```
dustmop: I believe that we shouldn't allow this. It breaks one of our earlier stated goals: explicit presentation of dependencies. In fact, I think that all arguments passed to "load_dataset" must be literal strings. I don't know if starlark allows us to check this, perhaps we need to parse the source ourselves?

text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+The former example is making use of a _priveledged resource_ (the local qri node), while the latter is a convenience function for creating legal string names. For this assumption to hold, priveledged resources must _only_ be available within special functions, an assumption we already enforce.
```
dustmop: Spelling: priveledged -> privileged

text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Name Resolution
+When the starlark execution environment encounters a `load_dataset` statement, it has to _resolve_ the name, connecting the string reference to a dataset object. This may or may not require network access, depending on if the user currently has all of the specified versions locally. All naming duties required by `load_dataset` will be offloaded to Qri's name resolution system, but a few interactions merit calling out here:
```
dustmop: Specify that if a dataset cannot be loaded, or doesn't exist anywhere, bail out and exit the script. There shouldn't be a recovery mechanism, which would allow probing datasets to see whether they exist or not.

text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Don't pin dependencies by default
+The default behaviour should be to **not pin dependencies**. We should add a new flag to `save` and `update` commands: `--pin-deps` that pins dependencies locally. The reason fro this choice relies on default settings in other parts of Qri. A "stock" installation of Qri defaults to storing datasets in an IPFS repository restricted to 10Gigs of space. In practice most datasets so far have been less than 100Mb in size, and most users coming to Qri either understand how to manually manage their IPFS storage, or are only using IPFS through Qri. These choices accumulate to lots of free space in the 10Gig repository that can go for some time before garbage collection is needed. If garbage collection _is_ required, the first thing that should go are transient dependencies required by previous transform scripts, which will be the case here. Future tools that perform graph analysis on versioned dependency trees can be introduced to add fine-grained pinning review & management. Did I mention that this work never stops?
```
dustmop: Spelling: fro -> for

text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,249 @@
...
+We're taking a lot of inspiration for the security model for imports & dependencies from the [deno project](https://github.com/denoland/deno)
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+### Loading from history
+One aspect not yet covered by this RFC is how to load previous versions of the dataset a script is affecting. The closest I have so far is supplying some sort of relative ref with no name:
+
+```python
+prev = load_dataset("~1")
```
dustmop: I don't think we should do this. At least, not without adding a solution for permissions / privilege escalation. It's not safe to assume that "~1" is a dataset meant for public consumption.

text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,249 @@
+- Feature Name: starlark load_dataset
+- Start Date: 2018-03-15
+- RFC PR: <!-- (leave this empty) -->
+- Issue: <!-- (leave this empty) -->
+
+# Summary
+[summary]: #summary
+
+Introduce a global function `load_dataset` to starlark scripts both loads the dataset and declares it as a dependency. Deprecate all `qri.load_dataset` functions, making `load_dataset` the only way to depend on an external dataset.
```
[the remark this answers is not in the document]
b5: delightful summary. Thanks for getting this on the record 😄

text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,249 @@
...
+
+def download(ctx):
+  fhv.get_body() # error: cannot access 'fhv' during download step
+
+def transform(ds,ctx):
+  fhv.get_body() # this will work
+```
+
+I now realize the problem is not in how the variable is declared, it is that the value stored at `fhv` is _stateful_. How the dataset is named is an orthagonal problem. The return-value style wins out in my mind on account of reading more clearly.
+
```
ramfox: I'm having a hard time understanding what this section is trying to express. 

Regardless of which import syntax we use, wouldn't the variable `fhv` have the same restrictions: `fhv.body()` would be able to be used in the transform step, but not the download step? 

Are you saying that by using a return-value syntax, we make it more clear that there is an object floating around that has special permissions/restrictions? Where the `load_dataset()` format doesn't make this as clear?

text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,249 @@
...
+One aspect not yet covered by this RFC is how to load previous versions of the dataset a script is affecting. The closest I have so far is supplying some sort of relative ref with no name:
+
+```python
+prev = load_dataset("~1")
+``` 
+
+This clearly needs more thought, I'm hoping to cover this in a subsequent RFC that helps refine & clarify the distinction between _names_ & _selection_.
+
+### Dataset object
+This RFC adds pressure to the api defined around dataset documents in starlark. We're on a collision course with a `dataframe-like api` that'll need to get worked out in a separate RFC. I think we can survive on the current model until a subsequent RFC can be written with extensive research dataframe-like APIs.
```
ramfox: grammar:
"can be written with extensive research dataframe-like APIs"
=>
"can be written with extensive research **on** dataframe-like APIs"

text/0023-starlark_load_dataset.md
b5: Yeah I should re-work this section to remove the intermediate thinking ("I now realize the problem is...")

You have the point right exactly, the import syntax _has no bearing_ on the behavior of the returned value, and that's the only I'm trying to make.

 Here it's getting confused with a second idea: that `fhv` can behave differently during different phases of a transform. I describe this as being _stateful_, but that's a separate point. from the point I'm trying to make. 

I think the answer is to remove this section altogether, and instead just point out that the returned `fhv` value will be stateful, and it's behaviour will be dictated by the `dataset` object, which will be defined in a separate RFC.

text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,249 @@
...
+- Start Date: 2018-03-15
+- RFC PR: <!-- (leave this empty) -->
+- Issue: <!-- (leave this empty) -->
+
+# Summary
+[summary]: #summary
+
+Introduce a global function `load_dataset` to starlark scripts both loads the dataset and declares it as a dependency. Deprecate all `qri.load_dataset` functions, making `load_dataset` the only way to depend on an external dataset.
+
+This RFC assumes all datasets are public. In the future `load_dataset` will be expanded, accepting optional arguments to scope the kinds of access a script is requesting.
```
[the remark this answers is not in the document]
b5: not totally true. I can write transforms against _my own_ unlisted datasets, just not against a dataset who's name I can't resolve.

text/0023-starlark_load_dataset.md
b5: agreed. Not sure how we're going to pull this off, but let's make it a design goal & see how hard starlark makes our lives.

b5:
Thanks for the solid review team, I've worked in requested changes. Can I get a vote on ratifying this?

text/0023-starlark_load_dataset.md
```diff
@@ -222,12 +213,12 @@ One aspect not yet covered by this RFC is how to load previous versions of the d
...
 prev = load_dataset("~1")
-``` 
+```
 
-This clearly needs more thought, I'm hoping to cover this in a subsequent RFC that helps refine & clarify the distinction between _names_ & _selection_.
+The current thinking is to _not_ do this because it's not safe to assume that "~1" is a dataset meant for public consumption. Loading historical entries clearly needs more thought, I'm hoping to cover this in a subsequent RFC that helps refine & clarify the distinction between _names_ & _selection_.
 
 ### Dataset object
-This RFC adds pressure to the api defined around dataset documents in starlark. We're on a collision course with a `dataframe-like api` that'll need to get worked out in a separate RFC. I think we can survive on the current model until a subsequent RFC can be written with extensive research dataframe-like APIs.
+This RFC adds pressure to the api defined around dataset documents in starlark. We're on a collision course with a `dataframe-like api` that'll need to get worked out in a separate RFC. I think we can survive on the current model until a subsequent RFC can be written with extensive research ondataframe-like APIs.
```
dustmop: Spelling, I think? "ondataframe-like" -> "on dataframe-like"

b5: lol nice catch

b5:
@ramfox can we get a review on this? thx!

text/0023-starlark_load_dataset.md
```diff
@@ -0,0 +1,240 @@
...
+[summary]: #summary
+
+Introduce a global function `load_dataset` to starlark scripts that both loads the dataset and declares it as a dependency. Deprecate all `qri.load_dataset` functions, making `load_dataset` the only way to depend on an external dataset.
+
+This RFC assumes all datasets are public, and that names refer to a dataset the script author can resolve. In the future `load_dataset` will be expanded, accepting optional arguments to scope the kinds of access a script is requesting.
+
+# Motivation
+[motivation]: #motivation
+
+Qri's goal is to build an environment where datasets are as a first class citizen. Qri currently monitors any call to `qri.load_dataset` and records the loaded dataset as a dependency. This results in a complete dependency graph.
```
ramfox: typo:
"where datasets are as a first" => "where datasets are a first

PR #40, opened 2019-03-26
RFC: merkle-derived messages

b5:


b5:
Closing on account of staleness. I still believe in the concept though, and hope to revisit someday

PR #41, opened 2019-03-26
RFC: Stats Component

ramfox:


text/0023-stats-component.md
```diff
@@ -0,0 +1,345 @@
...
+To start, Qri can create a default stats component on tabular (two dimentional) data. Because we want the cost of creating a stats component to be low, we only want to generate stats that can be calculated on a rolling basis.
+
+The different types of data that we can generate stats on are number (including integer and floating point), string, and boolean.
+
+We can generate different stats depending on the type of that column.
+
+Number:
+  valid - the number of cells in that column that contain valid numbers
+  error - the number of cells in that column that contain invalid numbers
+  missing - the number of cells in that column that do not contain data
```
ramfox: I added valid/error/missing, rather then just count after seeing in in kaggle. Feels like error is especially meaningful when we are trying to calculate stats, since values that error will not actually make into something like an average. 
We could remove missing and just use valid/error if we want to save space.

text/0023-stats-component.md
```diff
@@ -0,0 +1,345 @@
...
+
+Boolean
+  valid - the number of cells in that column that contain valid boolean values
+  error - the number of cells in that column that contain invalid boolean values
+  missing - the number of cells in that column that do not contain data 
+  trueCount - the number of cells that contain a true value
+  falseCount - the number of cells that contain a false value
+
+The structure of the default stats component depends on the structure of the data itself. If each row of the dataset body is expressed as an array, the stats component will be an array of objects. If each row of the dataset body is expressed as an object, the stats component will be an object of objects. One important thing to note, we must have a structure with a schema that describes the title and the type in order to create stats.
+
```
ramfox: So "One important thing to note, we must have a structure with a schema that describes the title and the type in order to create stats."... I don't actually know if this is true, but it feels like it needs to be. How else would we know at the start exactly what we are looking for when calculating the stats?

text/0023-stats-component.md
```diff
@@ -0,0 +1,345 @@
...
+    "type":"number",
+    "count": 4,
+    "min": 17,
+    "max": 22,
+    "avg": 19.5,    
+  },
+  "can_accept_more_students": {
+    "title":"can accept more students",
+    "type":"boolean",
+    "count": 4,
```
ramfox: just to point out, I kept 'count' here, rather then 'valid', 'error', and 'missing'

waiting for feedback on that before I standardize!

text/0023-stats-component.md
```diff
@@ -0,0 +1,425 @@
...
+    "trueCount": 2,
+    "falseCount": 2,
+  }
+}
+```
+
+In summary, qri can generate some basic stats on a dataset body, if the body is tabular. We also need a structure with a schema in order to understand what stats to be calculating.
+
+## high dimentional data
+Not only does 3 dimentinal (or higher) data have a less apparent structure, it is also diffcult to reason about the best way to display stats for 3 dimentional data.
```
b5: This section on high dimensional data feels less worked out. I think we should drop it for now, save the writing, and work on it in a follow-up RFC. Instead I think "high dimensional data" should just focus on stubbing out _how high dimensional statistics can be expressed_

> Calculating statistics on high dimensional data is out of scope for this RFC. however we do provide a way to arrange statistical metadata into hierarchies by nesting stats object with a key: `stats`. Given a sample of high-dimensional data: 

```
[
    {
      "type": "Feature",
      "properties": {
        "name":"Brooklyn"
      },
      "geometry": {
        "type": "Point",
        "coordinates": [
          -74.00,
          40.63
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {
        "name":"Manhattan"
      },
      "geometry": {
        "type": "Point",
        "coordinates": [
          -74.00,
          40.71
        ]
      }
    }
  ]
```

Here's an example of "stats that contain stats"

```js
{
  // other dataset components ...
  "stats":  [
    { "title": "type", "type" : "string" /* same as 2D string stat */ },
    {
       "title": "geometry",
       "type" : "object",
        "error" : 0,
        "valid" : 2,
        "missing" : 0, 
        "stats": {
          "geometry" : { "type" : "string" /* same as 2D string stat */ } ,
          "coordinates": {
             "type" : "geopoint",
             "avg" : [-74.00, 41.00]
          }
        }
    }
  }
}
```

What's important to note here is stats is a _recursive data structure_ that uses the keyword "stats" to build up statistical hierarchies. The actual calculation of these statistics is out of scope, but their presentation can be validated, mainly by reserving the `stats` keyword within a stats object.

b5:
We ended up implementing `stats` as a command instead of a component, largely thanks to the work put fourth on this RFC.

PR #42, opened 2019-04-23
RFC: cron

b5:
oook, the more I think about this, the bigger it gets. I'd love feedback. @rgardaphe I've tagged you b/c I'm trying out a new "term of art": _fog service_. Aside from the usual technical vetting, I'd love to hear other's thoughts

### DIFF
--- a/text/0024-cron.md
+++ b/text/0024-cron.md
@@ -1,0 +1,196 @@
+- Feature Name: cron
+- Start Date: 2019-04-12
+- RFC PR: <!-- (leave this empty) -->
+- Issue: <!-- (leave this empty) -->
+
+# Summary
+[summary]: #summary
+
+Cron generates new dataset verions with a damonized service that schedules & executes updates. Using cron, qri can keep data up to date on behalf of a user, and provide notifications of significant changes to datasets.
+
+The goals of this RFC are twofold:
+* define `qri cron`, a service for automating dataset updates
+* define the characteristics & requirements of running a "fog service"
+
+# Motivation
+[motivation]: #motivation
+
+As of this writing, all of my datasets are stale. They haven't been updated because, frankly, I'm too busy/distracted to remember to update them. This is a more serious issue than it seems at first. To put things bluntly: every dataset I create with Qri means more work for me, not less. This property has gotten in the way of me creating & maintaining lots of datasets. Since our goal is to convince others that there is merit in us working together on data, this makes getting to the "collaboration" part hard. 
+
+To keep datasets fresh, qri needs to automate the update process.  By _automate_ I mean: execute dataset updates as a background process, registered with the local operating system, according to some schedule I can specify. My computer should do the work of keeping datasets up to date for me, and should show me a log of what's been done for me, calling attention to significant changes.
+
+The magnitude of this shift to automation is noticeable. cron is _doing work for me_. It takes the process of dataset updating out of the categroy of chores, and opens up the possibliity that Qri can surprise me with new insights. 
+
+Adding cron starts to transition qri into territory best described as a _fog service_: a decentralized network of machines automatically producing local resources for a user that can aggregate to a broader network. Over time I'm hoping we can do more and more things under this "fog service" paradigm.
+
+Building a fog service requires special consideration to get right. This feature proposes running parts of Qri as a background process on a user's primary computing device. In this setting, we need to take great care to avoid interrupting a user's day-to-day computing experience, while also handling changing connectivity & resource availability. There is a very real chance that as a consequence of a cron job a user scheulded 6 months ago, Qri will be scheduled to start an intense script on a laptop with 5% battery left at a coffee shop while connected to restricted Wifi, while the user is nervously trying to get some dockerized python script to process a report that's due in an hour. We should aim to shape both the technical foundations and user experince of Qri to do the right thing in that context.
+
+This is a very high bar, but I think it's the standard we should work toward. Qri cron is both a big first step and a chance to iterate on meeting these requirements before including more parts of the Qri daemon in this "fog service" paradigm.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Here's a typical workflow for qri cron using datasets. This example shows working with the command line. As usual all of this would also be accessible over the local API. To start with, `qri cron` is a command with a number of subcommands. Ths first is `qri cron add`:
+
+```
+$ qri cron add me/dataset_name
+scheduled me/dataset_name
+updates every 2 days
+next update: Tues. April 11, 11:30:00
+
+WARNING: cron daemon is not currently running. start with:
+$ qri cron start
+to start the qri cron service. until then datasets will not auto-update.
+```
+
+In this case we've run `qri cron add` on a dataset that contains a valid `meta.accrualPeriodicity` field (in this case `R/P2D`). Qri used this value to infer a default update periodicity. If no `meta.accrualPeriodicity` were set, Qri would have returned an error:
+
+```
+error: no periodicity specified, please provide a periodicity argument. For example, use:
+$ qri cron add me/dataset_name R/P1W
+To an update every week. 
+More info at https://qri.io/docs/cron#periodicity
+```
+
+We can use this same technique to override the dataset-provided periodicity, in this case setting updates to every 8 hours, and use the `--publish` flag to have cron publish to the registry on successful update:
+```
+$ qri cron add me/dataset_name R/P8H --publish
+  scheduled: every 8 hours
+  next: Tues. April 11, 11:30:00
+  4 hours from now
+```
+In the above example we're providing two arguments to `qri cron add`: the path to a shell script, followed by an [ISO 8601 Repeating Interval string](https://en.wikipedia.org/wiki/ISO_8601#Repeating_intervals) that specifies unbounded updates every 8 hours. 
+
+Running `qri cron add` for the same dataset name `b5/dataset_name` will overwrite the previous job entry with new values. cron jobs establish their next execution date by examining the more recent of two time sources:
+1. last cron job execution
+2. dataset commit timestamp
+
+In this case a cron job has never run, so the scheduled date is derived from the commit timestamp (in this example the dataset was committed 4 hours ago). Any time either a new dataset version is created or the job is run, the next cron execution time is re-evaluated & scheduled. This way if a user manually runs `qri update me/dataset_name`, cron will reset to use that time as an interval basis, avoiding excessive updates.
+
+The flags for `qri cron add` are be a subset of those available for `qri update`. Anything `qri update` can do, cron should work the same way: 
+```
+$ qri update me/github_stars --config=org,qri-io,repo,qri-io --secrets=token,$GITHUB_TOKEN
+$ qri cron add me/github_stars --config=org,qri-io,repo,qri-io --secrets=token,$GITHUB_TOKEN
+```
+
+We can see scheduled jobs with `qri cron list`:
+```
+$ qri cron list
+1. b5/dataset_name
+   scheduled: every 8 hours
+   next: Tues. April 11, 11:30:00
+   4 hours from now
+```
+The conventions for `list` should follow all other "list command conventions" for pagination, but listing order should be chronological according to next scheduled update soonest update first, fathest update last.
+
+We can start the qri cron service like this:
+```
+$ qri cron start
+cron update service started. stop it with:
+$ qri cron stop
+```
+
+If the current operating system doesn't yet support a daemonized service, the deamon should start in-process and show a warning message explaining to run this command with some equivalent of `nohup`. We should work to make as few people as possible see that message. As shown in the dialog, the cron daemon must be stoppable with `qri cron stop`.
+
+With cron running Qri will run a lightweight background loop that checks for update events and executes them, logging the results to the event log. To see a log of cron events, run:
+```
+$ qri cron log
+* me/dataset_name
+  succeeded 3h4m ago
+* me/github_stats
+  failed 2d16h27m ago
+  message: error: network timeout
+ # ...
+```
+
+This will show a list of successful & failed job executions
+
+### Scheduling shell scripts
+
+Scheduling datasets directly is the ideal workflow in terms of portability, but a whole set of use cases open by adding the capacity to schedule & execute shell scripts within the same cron environment. While most of this is obviously available by directly using an operating system `cron` service, having scripts for dataset updates collected in one place makes scheduling & maintenance easier. For that reason, shell scripts should be supported as a first-class citizen:
+
+```
+$ qri cron add path/to/update_script.sh R/PT4H/2019-06-01T00:00:00Z
+```
+
+In this case we're case scheduling a cron job to run every 4 hours from now until June 1st, 2019. When qri encounters a file ending in a shell script extension (`.sh` on 'nix platforms), it copies the file into a directory within the fs repo and names the job according the the filename. `qri cron list` now shows both tasks, ordered by which script is the next to execute:
+
+```
+$ qri cron list
+1. update_script.sh
+   scheduled: every 1 hour
+   next: Tuesday April 11, 7:30:00
+   1 hour from now
+2. me/dataset_name
+   scheduled: every 8 hours
+   next: Tues. April 11, 11:30:00
+   4 hours from now
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### OS Support
+`qri cron start` registers a deamonized service with the host Operatring System. We should prioritize Platform support in the following order:
+1. darwin
+2. unix/linux
+3. windows
+
+### Determining Resource requirements, Monitoring & Constraining Resource Consumption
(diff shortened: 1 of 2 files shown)


text/0024-cron.md
```diff
@@ -0,0 +1,196 @@
+- Feature Name: cron
+- Start Date: 2019-04-12
+- RFC PR: <!-- (leave this empty) -->
+- Issue: <!-- (leave this empty) -->
+
+# Summary
+[summary]: #summary
+
+Cron generates new dataset verions with a damonized service that schedules & executes updates. Using cron, qri can keep data up to date on behalf of a user, and provide notifications of significant changes to datasets.
```
ramfox: "damonized" -> "daemonized"

text/0024-cron.md
```diff
@@ -0,0 +1,196 @@
...
+
+The goals of this RFC are twofold:
+* define `qri cron`, a service for automating dataset updates
+* define the characteristics & requirements of running a "fog service"
```
ramfox: "fog service" is a good abstraction of what's happening, but I wonder if the word "fog" has too many negative associations. I like that it's akin to "cloud" but literally closer to earth. But it can be interpreted as "confusing" or "shady" or "opaque"

text/0024-cron.md
dustmop: verions -> versions

text/0024-cron.md
```diff
@@ -0,0 +1,196 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+As of this writing, all of my datasets are stale. They haven't been updated because, frankly, I'm too busy/distracted to remember to update them. This is a more serious issue than it seems at first. To put things bluntly: every dataset I create with Qri means more work for me, not less. This property has gotten in the way of me creating & maintaining lots of datasets. Since our goal is to convince others that there is merit in us working together on data, this makes getting to the "collaboration" part hard. 
+
+To keep datasets fresh, qri needs to automate the update process.  By _automate_ I mean: execute dataset updates as a background process, registered with the local operating system, according to some schedule I can specify. My computer should do the work of keeping datasets up to date for me, and should show me a log of what's been done for me, calling attention to significant changes.
+
+The magnitude of this shift to automation is noticeable. cron is _doing work for me_. It takes the process of dataset updating out of the categroy of chores, and opens up the possibliity that Qri can surprise me with new insights.
```
dustmop: categroy -> category

text/0024-cron.md
```diff
@@ -0,0 +1,196 @@
...
+scheduled me/dataset_name
+updates every 2 days
+next update: Tues. April 11, 11:30:00
+
+WARNING: cron daemon is not currently running. start with:
+$ qri cron start
+to start the qri cron service. until then datasets will not auto-update.
+```
+
+In this case we've run `qri cron add` on a dataset that contains a valid `meta.accrualPeriodicity` field (in this case `R/P2D`). Qri used this value to infer a default update periodicity. If no `meta.accrualPeriodicity` were set, Qri would have returned an error:
```
dustmop: I find R/P hard to mentally process. Is this a standard syntax used somewhere I'm not familiar with? How about making it a flag like `--when 2D`.

text/0024-cron.md
```diff
@@ -0,0 +1,196 @@
...
+The conventions for `list` should follow all other "list command conventions" for pagination, but listing order should be chronological according to next scheduled update soonest update first, fathest update last.
+
+We can start the qri cron service like this:
+```
+$ qri cron start
+cron update service started. stop it with:
+$ qri cron stop
+```
+
+If the current operating system doesn't yet support a daemonized service, the deamon should start in-process and show a warning message explaining to run this command with some equivalent of `nohup`. We should work to make as few people as possible see that message. As shown in the dialog, the cron daemon must be stoppable with `qri cron stop`.
```
dustmop: deamon -> daemon

text/0024-cron.md
```diff
@@ -0,0 +1,196 @@
...
+   scheduled: every 8 hours
+   next: Tues. April 11, 11:30:00
+   4 hours from now
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### OS Support
+`qri cron start` registers a deamonized service with the host Operatring System. We should prioritize Platform support in the following order:
```
dustmop: Operatring -> Operating

text/0024-cron.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Scheduling shell scripts
+
+Scheduling datasets directly is the ideal workflow in terms of portability, but a whole set of use cases open by adding the capacity to schedule & execute shell scripts within the same cron environment. While most of this is obviously available by directly using an operating system `cron` service, having scripts for dataset updates collected in one place makes scheduling & maintenance easier. For that reason, shell scripts should be supported as a first-class citizen:
+
+```
+$ qri cron add path/to/update_script.sh R/PT4H/2019-06-01T00:00:00Z
+```
+
+In this case we're case scheduling a cron job to run every 4 hours from now until June 1st, 2019. When qri encounters a file ending in a shell script extension (`.sh` on 'nix platforms), it copies the file into a directory within the fs repo and names the job according the the filename. `qri cron list` now shows both tasks, ordered by which script is the next to execute:
```
dustmop: Hate to be that guy, but any thoughts as to how this will work on Windows? Do we require a cygwin / mingw / msys environment? Do we support powershell? Do we support batch (lol, no, not doing that)?

text/0024-cron.md
```diff
@@ -0,0 +1,196 @@
...
+  failed 2d16h27m ago
+  message: error: network timeout
+ # ...
+```
+
+This will show a list of successful & failed job executions
+
+### Scheduling shell scripts
+
+Scheduling datasets directly is the ideal workflow in terms of portability, but a whole set of use cases open by adding the capacity to schedule & execute shell scripts within the same cron environment. While most of this is obviously available by directly using an operating system `cron` service, having scripts for dataset updates collected in one place makes scheduling & maintenance easier. For that reason, shell scripts should be supported as a first-class citizen:
```
dustmop: Alternative implementation: add support for "hooks", like git has. Then, I could choose to run a shell script only when `cron` has found new data.

Related: do shell scripts have a one-to-one relationship with datasets, or do we support a shell script updates multiple datasets?

text/0024-cron.md
```diff
@@ -0,0 +1,196 @@
...
+
+This is a very high bar, but I think it's the standard we should work toward. Qri cron is both a big first step and a chance to iterate on meeting these requirements before including more parts of the Qri daemon in this "fog service" paradigm.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Here's a typical workflow for qri cron using datasets. This example shows working with the command line. As usual all of this would also be accessible over the local API. To start with, `qri cron` is a command with a number of subcommands. Ths first is `qri cron add`:
```
dustmop: Larger question: why keep a division between `qri cron` and `qri update`? Are they different enough that they need to be separate commands? Is it worth having to explain to users why there are two commands that do nearly the same thing? I'm reminded of our long ago discussions around `qri save` and `qri new`.

Alternatively, could `qri cron` be merged into a subset of `qri update`, like `qri update schedule me/dataset_name`? Doing so would also help solve the dependency problem, wherein cron invokes update, but update has to notify cron when run manually.

Related but separate to this question, do we want `qri cron` (whether or not it gets combined with update) to be able to fetch new data from other peers, the way that `qri update` can?

text/0024-cron.md
b5: Personally, I'm ok with the "confusing" side of things. I'm hoping we can stay lighthearted about the whole thing. To me the "abstraction phrase" needs to emphasize honesty over wanting to be placed in a positive light

text/0024-cron.md
b5: Never don't be that guy. I have literally zero idea, and we should really figure it out. I think what matters here is making sure this kind of thing is _possible_ on windows, even if we don't have a solid plan. As our foremost windows expert, do you think this is doable if we pick a lane in terms of environment?

text/0024-cron.md
b5: OooOOOh hooks is a very interesting concept to bring up while we're having this conversation.

Many use cases will create new versions as the _result_ of running a script, not just a consequence of a dataset update completing. Concrete examples with high-level steps for completion:

#### libp2p_node_count
* port forward our cloud-based prometheus server to my local machine
* run `qri update me/lip2p_node_count`
* close the ssh tunnel

#### edgi_webmon_sitemap
* spin up a copy of `walk` with a preconfigured job file that creates a sitemap
* wait a number of hours
* once the process exits it'll write a file called `sitemap.json`
* run `qri save --body sitemap.json me/some_sitemap`

In both cases we're using a shell script to generate data, and then adding it to Qri, so the script should come first.

That being said, the question of how hooks might intersect with this conversation should be addressed. We've also had requests for _webhooks_, both of which are basically "run this when x event happens".

To me the mental model here is "scheduling" or "orchestration". cron is pulling Qri's puppet strings, whereas hooks are reactionary, and should be build as a separate service that tails additions to the qri event log.

> do shell scripts have a one-to-one relationship with datasets, or do we support a shell script updates multiple datasets?

To me there is no correlation at all. I think it's totally fine to run a script that doesn't update a qri dataset at all, so it's a zero-or-more relationship. shell scripts can and should call `qri update` directly, and if they do so multiple times, that's fine.

In documentation I think we would _strongly_ recommend against running multiple updates in the same script, as it's no longer conceptually transactional. You'd have to roll back a partially-successful script by hand. Thankfully there's a full audit trail for this sort of thing.

text/0024-cron.md
b5: >  do we want qri cron to be able to fetch new data from other peers, the way that qri update can?

Yes. Cron schedules and automates the qri update process, so whatever behavior you would have running `qri update` directly should be the exact same here. If `connect` is running when `udpate` is called, the d.web is available and can be used. If not, more requests fall back to the registry. This is why we really want to move toward `qrid` quickly to get easier dweb "toggling".

** ** 

>  why keep a division between qri cron and qri update?

really, really good question. @dustmop & I chatted about this a bunch offline, and agreed we see no reason to keep both. Like the save/new debate, it's confusing to have the concept of updating beside the concept of scheduling.

A few really great points came out of that discussion:

### This RFC should be a breaking change on update
`cron` isn't a very user friendly. `update` is a lot easier, so we agreed this rfc should switch to modifying the way update works instead of introducing a new term.

### the current `qri update` is a lot like the refresh button on your browser
`qri update` as it currently works is just an alias for `qri save --recall tf`, which in english is just "re-run that transform script and create a new version if something changes" . That sounds a lot like "refresh". Now that we're thinking about scheduled fetching, think about how often you hit the refresh button on your browser.

In this metaphor it's clear that scheduled updates are the default. When data is syncing to an external source, qri should be doing the sync work for you, and hand you a button that works a lot like "refresh", that lets you override qri's default for an immediate check. Dustin used the metaphor of when you've just signed up for a service, and you're refreshing your email inbox because you _know_ an email is coming.

One issue we can't really get around is "default" here only applies in situations where updates can be automated. There are lots of workflows in qri that don't work this way. This means we'll always need some sort of "automation setup" step.

text/0024-cron.md
b5: Yeah, it's [iso8601](https://en.wikipedia.org/wiki/ISO_8601#Repeating_intervals), which many metadata standards use. That being said, we don't _need_ to stick with it. Do you have other specs/standards we should consider?

b5:
While working on the proof-of-concept code after the `qri update` refactor, I've realized that this should also work as part of this RFC:

```
qri update schedule --body path/to/file.csv me/dataset_name R/P1D
```

meaning we can schedule _manual_ transforms as well as scripted transforms. This opens the door to UI that lets a user drag an excel file over the frontend and "schedule routine saves" that automatically creates new versions of a file. Combine this with dropbox-style directory syncing and we have a really nice collaboration & audit workflow.

The downside: if the filepath ever changes, scheduled updates will break. I think it's worth it, given that with this users can create "data feeds" from an excel file without writing a line of code.

text/0024-scheduled-updates.md
```diff
@@ -19,104 +19,109 @@ As of this writing, all of my datasets are stale. They haven't been updated beca
...
+$ qri update schedule me/dataset_name
 scheduled me/dataset_name
 updates every 2 days
 next update: Tues. April 11, 11:30:00
 
-WARNING: cron daemon is not currently running. start with:
-$ qri cron start
-to start the qri cron service. until then datasets will not auto-update.
+WARNING: qri update is not currently running. start with:
+$ qri update start
```
dustmop: I like all of the new names, except for `qri update start`. I don't think it strongly enough communicates that its launching a long-lived operation. With services, like in systemd and launchd, `<some-service> start` is obviously long-lived because it's talking just about a service. But `update` does many different things. `qri update start` is very close to `qri update run`. I guess what I'm getting at is that "service" should be somewhere in the name, like `qri update start-service'.

text/0024-cron.md
dustmop: No, as long as there's a source of inspiration I'm good with it, I was unaware of this.

text/0024-scheduled-updates.md
b5: agreed. I'm thinking `qri update service start`

b5:
Ok, changes made. re-requesting for final review

text/0024-scheduled-updates.md
```diff
@@ -0,0 +1,214 @@
...
+
+In the above example we're providing two arguments to `qri update schedule`: a dataset reference, followed by an [ISO 8601 Repeating Interval string](https://en.wikipedia.org/wiki/ISO_8601#Repeating_intervals) that specifies unbounded updates every 8 hours. 
+
+Running `qri update schedule` for the same dataset name `b5/dataset_name` will overwrite the previous job entry with new values. update jobs establish their next execution date by examining the more recent of two time sources:
+1. last update execution
+2. dataset commit timestamp
+
+In this case an update has never run, so the scheduled date is derived from the commit timestamp (in this example the dataset was committed 4 hours ago). Any time either a new dataset version is created or the update is run, the next update execution time is re-evaluated & scheduled. This way if a user manually runs `qri update run me/dataset_name`, update will reset to use that time as an interval basis, avoiding excessive updates.
+
+The flags for `qri update schedule` are be a subset of those available for `qri save`. Anything `qri save` can do, update should work the same way. For example:
```
ramfox: "are be a subset" => "are a subset"

text/0024-scheduled-updates.md
```diff
@@ -0,0 +1,214 @@
...
+We can start the qri update service like this:
+```
+$ qri update service start
+cron update service started. stop it with:
+$ qri update service stop
+```
+
+If the current operating system doesn't yet support a daemonized service, the deamon should start in-process and show a warning message explaining to run this command with some equivalent of `nohup`. We should work to make as few people as possible see that message. As shown in the dialog, the update daemon must be stoppable with `qri update stop`.
+
+With cron running Qri will run a lightweight background loop that checks for update events and executes them, logging the results to the event log. To see a log of cron events, run:
```
ramfox: "With cron running Qri will run a" => "With cron running, Qri will run a"

text/0024-scheduled-updates.md
```diff
@@ -0,0 +1,214 @@
...
+
+### better error handling, classification and reporting
+Debugging _why a cron job didn't succeed_ is going to come entirely to error messages. `error: i/o timeout` isn't enough to tell a user "hey when this cron job ran you didn't have an internet connection, and your transform script makes HTTP requests". It should come as close to that as possible. I'm hoping we can dedicate dev cycles in the future to improving our error message story, ideally integrate with [upcoming changes to errors in go](https://go.googlesource.com/proposal/+/master/design/29934-error-values.md).
+
+### Multi-Tenancy
+A user may be running multiple qri repositories, each with their own cron jobs. We'll need to come up with a way to deal with this in both the short and long term.
+
+### Cron log file persistence
+writing the output of a shell script that may print who-knows-what to a log file buried in a hidden folder sounds like a great way to eat up hard drive space & irritate users. We should figure out a ux for "cleanup", and think about weather "cleaning" should be a process that operates across all Qri serivces at once.
```
ramfox: "weather" => "whether"

text/0024-scheduled-updates.md
```diff
@@ -0,0 +1,214 @@
...
+This scheduled shell script has a zero-or-more relationship with datasets. This means it's possible to run a script that doesn't update a qri dataset at all, or have a single script update multiple datasets. To change datasets, scripts will need to call `qri save`, which will record updates as if they we manually invoked with the qri event log.
+
+In documentation we will _strongly_ recommend against running multiple updates in the same shell script, as such a script is no longer conceptually transactional. Users would have to roll back a partially-successful script by hand. Thankfully there's a full audit trail for this sort of thing. We'll allow this behaviour because there are a number of advanced use-cases where being able to perform multiple updates in the same script is useful, and because it's much easier to consider the shell script opaque.
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### OS Support
+`qri update start` registers a deamonized service with the host Operating System. We should prioritize Platform support in the following order:
```
dustmop: Change to `qri update service start`

PR #45, opened 2019-07-02
rfc(filesystem integration): couple qri datasets with a folder on the local file system

chriswhong:
Adds filesystem integration to qri datasets including a current working directory context for some qri commands.  

The goal is to provide an experience that is more analogous to the git workflow, where cloning produces files on the local filesystem, and working changes to files are sensed prior to committing a new version.

### DIFF
--- a/text/0000-filesystem-integration.md
+++ b/text/0000-filesystem-integration.md
@@ -1,0 +1,72 @@
+- Feature Name: Filesystem Integration
+- Start Date: 2019-07-02
+- RFC PR:
+- Issue:
+
+# Summary
+[summary]: #summary
+
+Adds filesystem integration to qri datasets including a current working directory context for some qri commands.  The goal is to provide an experience that is more analogous to the git workflow, where working changes to files are sensed prior to committing a new version.
+
+# Motivation
+[motivation]: #motivation
+
+The motivation for this feature is to provide a more tangible connection between the qri versioning/publishing ecosystem and the file-based data that data practitioners are used to working with. Since all of the information in a qri dataset maps to files, we should just make those files exist automatically when a qri dataset is "added" to the user's local machine.  Likewise, we should watch for changes to those files, and ensure that they are all in a valid state before allowing the user to commit changes to a qri dataset.
+
+*This will significantly lower the barrier to entry for new users to qri, especially GUI users, as they can quickly pull down published datasets and begin working with them without having to understand how IPFS/QRI stores and maintains the data.*  
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The intent is to create a workflow more analogous with the git workflow.  When a new user wants the files in a remote git repo, they use `git clone {repoUrl}` and a new directory is created on their filesystem containing those files.  They can use _whatever tooling they want_ to make changes to the files, and commit them locally when they reach a critical point.
+
+We want the qri dataset user to have a similar experience: `qri add /b5/world_bank_population/` should create a new directory called `world_bank_population` on the filesystem, containing the following files:
+
+- `body.csv`
+- `meta.json`
+- `structure.json`
+- `transform.star`
+
+To change the dataset, the user uses _whatever tooling they want_ to modify the files, then commits the changes.
+
+## Commands
+
+Under this proposal, `qri use` is no longer relevant, as qri commands should be run from the qri dataset's directory.
+
+- `qri add` - creates a copy of the dataset's files in the present working directory
+- `qri status` - provides a report of the current state of the files in the current directory versus the last commit to the dataset
+- `qri save` - commits changes on the current dataset, and should throw errors if the changes to be committed are not valid
+
+## Strict Filenames
+
+This feature allows for qri datasets to live anywhere on the user's filesystem, and be mixed-in with other files and folders.  The initial implementation should require the strict filenames and extensions listed above (with the exception that `body` can have extension csv, json, xlsx, etc)
+
+This means `qri status` and `qri save` should ignore any non-qri files in the dataset directory.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+- Something about the invisible `.qri-ref` file that links the current directory to the central qri store
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- More overhead around the core qri service
+- Permissions/Cross-platform filesystem issues?
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+This design is preferable because it makes Qri more relatable to data consumers who are used to working with data files.  It also emulates the established and popular workflow of git.
+
+If we don't do this, qri users have the additional burden of manually exporting qri components to files, and must use more advanced commands/tooling/scripting to commit changes to a dataset. _This proposal opens up the universe of qri-compatible tools to anything that can save json and csv files to the filesystem_
+
+This feature is also critical to the desired GUI for Qri, which would make the Qri frontend behave more like Github Desktop.  The Qri frontend will clone, validate commits, show diffs, allow for pushing and publishing, etc.  Some basic editing will be possible via the frontend, but most users will move to another environment to actually make changes.
+
+# Prior art
+[prior-art]: #prior-art
+
+Git is the inspiration for this model, specifically the idea that cloning a repository results in new files on the local machine that can be modified and subsequently committed as a new version.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
--- a/text/0000-filesystem-integration.md
+++ b/text/0000-filesystem-integration.md
@@ -60,7 +60,13 @@ Notes:
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-- Something about the invisible `.qri-ref` file that links the current directory to the central qri store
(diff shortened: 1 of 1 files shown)


text/0000-filesystem-integration.md
```diff
@@ -0,0 +1,72 @@
...
+- `structure.json`
+- `transform.star`
+
+To change the dataset, the user uses _whatever tooling they want_ to modify the files, then commits the changes.
+
+## Commands
+
+Under this proposal, `qri use` is no longer relevant, as qri commands should be run from the qri dataset's directory.
+
+- `qri add` - creates a copy of the dataset's files in the present working directory
```
b5: `qri add` is an already existing command for fetching a dataset from the network. `qri add ramfox/nyc_candidates` will grab the dataset from either the network or. We could consider renaming the current `qri add` to `qri fetch` or `qri download`. Considering we're talking about mapping closer to git, we should consider such a change anyway, or at least make sure that users who try to use `add` in the same way as git don't create problems for themselves.

But there's a larger point to keep in mind here: Qri has no _staging area_, and I think that's a good thing. Because we have a declarative data model, we don't need to ask what to add. The user controls what's staged by manipulating the contents of the directory. Files that don't match our naming conventions aren't included. This means we can skip this "add" step entirely.

text/0000-filesystem-integration.md
```diff
@@ -0,0 +1,72 @@
...
+
+*This will significantly lower the barrier to entry for new users to qri, especially GUI users, as they can quickly pull down published datasets and begin working with them without having to understand how IPFS/QRI stores and maintains the data.*  
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The intent is to create a workflow more analogous with the git workflow.  When a new user wants the files in a remote git repo, they use `git clone {repoUrl}` and a new directory is created on their filesystem containing those files.  They can use _whatever tooling they want_ to make changes to the files, and commit them locally when they reach a critical point.
```
b5: I don't think the intent is to be more analogous with git, but rather to _use a common interface_: the filesystem. This is analogous to git, which is nice, but there are subtle differences we'll need to avail users to.

I want to keep us focused on taking inspiration from the good bits of existing version control systems. We should aim to do better than git's workflow when it comes to versioning data

text/0000-filesystem-integration.md
```diff
@@ -0,0 +1,72 @@
...
+
+- `qri add` - creates a copy of the dataset's files in the present working directory
+- `qri status` - provides a report of the current state of the files in the current directory versus the last commit to the dataset
+- `qri save` - commits changes on the current dataset, and should throw errors if the changes to be committed are not valid
+
+## Strict Filenames
+
+This feature allows for qri datasets to live anywhere on the user's filesystem, and be mixed-in with other files and folders.  The initial implementation should require the strict filenames and extensions listed above (with the exception that `body` can have extension csv, json, xlsx, etc)
+
+This means `qri status` and `qri save` should ignore any non-qri files in the dataset directory.
```
b5: This is going to become a problem when we try to attach arbitrary files to a dataset

text/0000-filesystem-integration.md
```diff
@@ -0,0 +1,72 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- More overhead around the core qri service
+- Permissions/Cross-platform filesystem issues?
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+This design is preferable because it makes Qri more relatable to data consumers who are used to working with data files.  It also emulates the established and popular workflow of git.
```
b5: I like this positioning in relation to git more.

text/0000-filesystem-integration.md
```diff
@@ -0,0 +1,72 @@
...
+
+If we don't do this, qri users have the additional burden of manually exporting qri components to files, and must use more advanced commands/tooling/scripting to commit changes to a dataset. _This proposal opens up the universe of qri-compatible tools to anything that can save json and csv files to the filesystem_
```
b5: This could also have fun implications for `qri update`. One could setup `qri update` to auto-publish from the set directory

text/0000-filesystem-integration.md
```diff
@@ -0,0 +1,72 @@
...
+
+This feature is also critical to the desired GUI for Qri, which would make the Qri frontend behave more like Github Desktop.  The Qri frontend will clone, validate commits, show diffs, allow for pushing and publishing, etc.  Some basic editing will be possible via the frontend, but most users will move to another environment to actually make changes.
+
+# Prior art
+[prior-art]: #prior-art
+
+Git is the inspiration for this model, specifically the idea that cloning a repository results in new files on the local machine that can be modified and subsequently committed as a new version.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
```
b5: @dustmop brought this up earlier: we need to decide weather the relationship of working dataset directories is a 1-1 or a 1-many. Can I have multiple directories on my computer that point to the same dataset? What happens when I run save, do all these directories update? I'm inclined to say that for now we can only have _one_ working dataset directory per dataset. Unless others can think of a common use case where having multiple working dataset directories is useful.

Another thing we need to consider: "checkout", and the notion of a "dirty" working dataset directory. We'll likely need a `qri checkout` command that pulls data from a previous version into the working dataset directory

text/0000-filesystem-integration.md
```diff
@@ -0,0 +1,72 @@
...
+The intent is to create a workflow more analogous with the git workflow.  When a new user wants the files in a remote git repo, they use `git clone {repoUrl}` and a new directory is created on their filesystem containing those files.  They can use _whatever tooling they want_ to make changes to the files, and commit them locally when they reach a critical point.
+
+We want the qri dataset user to have a similar experience: `qri add /b5/world_bank_population/` should create a new directory called `world_bank_population` on the filesystem, containing the following files:
+
+- `body.csv`
+- `meta.json`
+- `structure.json`
+- `transform.star`
+
+To change the dataset, the user uses _whatever tooling they want_ to modify the files, then commits the changes.
```
b5: I ❤️ this very much.

text/0000-filesystem-integration.md
```diff
@@ -0,0 +1,72 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+The intent is to create a workflow more analogous with the git workflow.  When a new user wants the files in a remote git repo, they use `git clone {repoUrl}` and a new directory is created on their filesystem containing those files.  They can use _whatever tooling they want_ to make changes to the files, and commit them locally when they reach a critical point.
+
+We want the qri dataset user to have a similar experience: `qri add /b5/world_bank_population/` should create a new directory called `world_bank_population` on the filesystem, containing the following files:
+
+- `body.csv`
+- `meta.json`
+- `structure.json`
+- `transform.star`
```
b5: There's a long conversation to have about `dataset.yaml` here, I'll take some time to write out thoughts & post back here

text/0000-filesystem-integration.md
chriswhong: 👍 Added the bits about using a "common interface"

text/0000-filesystem-integration.md
chriswhong: Added the 1-1 rule as a start and a `checkout` command.

text/0000-filesystem-integration.md
```diff
@@ -0,0 +1,86 @@
...
+
+A subset of the full qri `dataset.json` with only user-editable keys: `meta`, `structure`, `transform`.  Everything else is managed by qri behind the scenes.
+
+### `body.[csv|json|xlsx]`
+
+The data file!
+
+## Commands
+
+- `qri init {name}` - creates an "empty" qri dataset in the user's local ipfs repo, adds a directory `/{name}` in the present working directory, links that directory to the qri dataset and populates an empty `dataset.json`
```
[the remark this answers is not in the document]
chriswhong: You're right, just as you `git init` in an already existing directory, `qri init` should not make a directory, it should just do the linking step.

I do think another command that makes the directory and the files in one step would be helpful.

text/0000-filesystem-integration.md
```diff
@@ -0,0 +1,86 @@
...
+
+- `qri clone` - hybrid command that adds an existing dataset to the user's local IPFS repo _AND_ creates a linked working directory on the filesystem.
```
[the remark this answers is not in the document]
chriswhong: This feels similar to `git remote set-url` to me, and I think `qri link` is the right way to go.  User could link a working directory to a different local dataset if they want to.

text/0000-filesystem-integration.md
```diff
@@ -0,0 +1,136 @@
...
+## Commands
+
+- `qri init` - interactive prompt that sets up a qri dataset with linked files in the current directory. (User should have already created a directory to hold the dataset, `cd` into it, then run `qri init`)
+
+This should have an `non-interactive` flag that only needs a dataset name, which will link an empty dataset working directory to a name reference in the qri repo. (reserving the name for a future commit)
+
+### Prompts:
+  -  **Dataset Name** - (autopopulate with the name of the current directory) generates `meta.json` in the working directory with the supplied name as `title` and boilerplate `description`, `keywords`, etc
+
+  -  **File Type** - csv/json? Based on the user's selection, add dummy `body.csv|json` and `schema.json` to the working directory.
```
dustmop: Is this RFC final? After our in person discussions yesterday, I wasn't sure if we ended on having a schema.json file created by default. Also, we *don't* create dataset.json, but a user can make their own for advanced options?

b5:
Ooook we're literally about to release code that implements this, and we've all agreed offline that FSI is super dope. _This RFC is hereby ratified_

PR #48, opened 2020-04-10
RFC: Apply command fixes how save, transform, FSI work

dustmop:


text/0026-transform_application.md
```diff
@@ -0,0 +1,107 @@
...
+
+Looking back on this, I don't think this was the best decision. We're affecting the core operation of save and the way it outputs information in order to support a tricky-to-use secondary feature. Furthermore, it's worth asking if this is the correct way to think about developing and introspecting transforms. For the dry-run case, we're confusing `save` which creates commits, with observing and processing the result of a dry-run, which is a read-only operation. The thing we actually care in this case is executing the transform and seeing it's output, not performing a `save`.
+
+At the same time, we have never had a full explanation for how transforms and FSI are supposed to work together, and there's lots of bugs now in how dry-run and transform interact with FSI. We're at a point where we need to re-evaluate how these pieces work together.
+
+Let's go back to the very beginning.
+
+## Manual Save
+
+In the early days, it was decided that the right way to think about a `save` operation on a dataset that already existed was as a "patch application". Let's say I have a dataset named "dustomp/my_ds" that looks like this:
```
feep, line 29: dustomp

feep:
Hey, thanks.

This helps me understand what’s going on beneath, and what the plan is.

I’ll probably have comments on implementation and UI/UX. They will probably come during... implementation.

Excited to see it happen.

text/0026-transform_application.md
```diff
@@ -0,0 +1,107 @@
...
+
+This transform changes the shape of the data, so it doesn't make sense to apply it again:
+![transform that changes shape](https://github.com/qri-io/rfcs/blob/rfc-xform/img/dataset_03_xform_diff_shape.png)
+
+This transform keeps the same shape of the data; running it a second time makes sense:
+![transform that maintains shape](https://github.com/qri-io/rfcs/blob/rfc-xform/img/dataset_04_xform_same_shape.png)
+
+Similar to a manual save, a transform script represents a state transition, with a reproducible method to execute that state transition, and a mechanism called "recall" to clone that state transition when applicable.
+
+A crucial fact about this description is that a manual save and a transform cannot both happen at the same time. They are both state transitions, and since we define a linear history of commits, it's illegal to have two transitions at once. Neither forking nor merging are allowed:
```
b5, line 57: Hrm... you _can_ mix manual & scripted transforms in the same commit. both just can't modify the same path.

text/0026-transform_application.md
```diff
@@ -0,0 +1,107 @@
...
+  * The stashed body is unchanged from the previous commit, replace it with the transforms' output
+  * Otherwise throw an error because the user may be making a mistake
+4) Not mentioned above, but once Desktop has better support for transform, we probably will eventually want some interactive development environment. Perhaps transforms can be incrementally compiled and applied, resulting in fast turnaround.
+
+To expand on this idea, applying a transform could be generalized out to other tasks that execute some sort of data pipeline. For example, if we bring back viz, generating those visuals could also be seen as a type of apply operation, with the script being some sort of imaging library, and the output being static images. Or, our `qri update` feature could be somehow incorporated into this. Any sort of task that computes using a dataset to produce some view or artifact could be folded into this command. Better to have a generalized tool than create a one-off command that solves only a narrow problem.
+
+More details of the implementation should wait until the feature is further discussed and some rough consensus can be reached.
```
b5, line 107: I think we should continue to rely on file extensions for syntax, and magic filenames for FSI integrations. 


What if `qri save` got an `apply` flag? Working outside of FSI that could look like this:

```
qri save --body body.csv --apply transform.star me/dataset
```

in FSI land a file named `transform.star` would be auto-applied. I could also choose to _not_ have a file called `transform.star` and instead  have lots of other scripts in my FSI-linked directory that I could override with `--apply`. `qri save --apply correct_my_math.star` feels pretty natural

text/0026-transform_application.md
```diff
@@ -0,0 +1,107 @@
...
+  * The stashed body matches the output of the transform
+  * The stashed body is unchanged from the previous commit, replace it with the transforms' output
+  * Otherwise throw an error because the user may be making a mistake
+4) Not mentioned above, but once Desktop has better support for transform, we probably will eventually want some interactive development environment. Perhaps transforms can be incrementally compiled and applied, resulting in fast turnaround.
+
+To expand on this idea, applying a transform could be generalized out to other tasks that execute some sort of data pipeline. For example, if we bring back viz, generating those visuals could also be seen as a type of apply operation, with the script being some sort of imaging library, and the output being static images. Or, our `qri update` feature could be somehow incorporated into this. Any sort of task that computes using a dataset to produce some view or artifact could be folded into this command. Better to have a generalized tool than create a one-off command that solves only a narrow problem.
```
b5, line 105: _Ugh this comment is taking too long to write! I'll try to come back & finish this_

I really like this, and feel like you've hit on a big concept here @dustmop. I think of `apply` as `execute scripts on datasets`. One could also say "apply transforms datasets". We can even get formal about it conceiving of as apply as executing a single function with one or more datasets as input, and returns a single function.

```
apply = fn(ds1, ds2, ...): ds 
```

Now, that function can _absolutely_ have side effects (making network calls is a side effect), Capturing as many of those side effects as possible in the resulting dataset

I totally think we should make the area `apply` covers large, extending it to all "executable input". Apply may be a 

We write lots of functions within the qri codebase itself that would qualify as "apply"

```
sql
diff
render
stats
validate
```

This is mainly so we can think about it formally, where apply

I'm not saying any of those

Arqu: I'd like to double down on this. Actually for the longer term I think it would be good to have all "modifier" operations be part of the apply ecosystem, governed by the "applicator engine" of qri. Ideally this would detach data storage/versioning from any other operations which would split the responsibility of qri on data management and data processing into distinct buckets. Hopefully this would allow for writing very custom toolchains later on top of the "apply" paradigm and potentially be exposed to the user as well to suit their custom needs.

dustmop: I think we need to constrain the definition of `apply` somewhat, so that it doesn't become simply synonymous with "code", and therefore lose all useful meaning. To do that, we need to figure out what's the benefit of defining `application`, what sort of guarantees do we want from it, as something that acts like running a transform, but is more general.

Suggestions: 
- `apply` uses a dataset as input, possibly has other inputs, and the output is something that can become a component of a dataset. So something like performance measurements wouldn't qualify. I don't think `validate` does either, but am not sure.
- `apply` is always optional, but is a separate code path from a normal save. This means `diff` doesn't qualify, because we actually use `diff` when making a commit, both to see if anything changed, and to calculate the friendly commit message.
- The others, `sql`, `render`, `stats` are all good by me.

Arqu: Hmm, I agree. Cramming everything into apply now just gets us back to the initial problem just at a different subcomand. 
I'm all for it to carve out a clear definition of apply and what can fall under it.

If I had to pick, your first suggestion of apply makes the most sense to me. And `validate`, unless the output is saved should not be an `apply` operator anyways, it's result is transient for the most part and think `diff` should also fall into this bracket. It offers valid and useful info but is not part of the dataset, it just describes the transition similar as to how `validate` describes the current state of the dataset.

b5: I'm in on having `validate` and `diff` _not_ be part of the apply family. Points made above make a great deal of sense. 

To that end, maybe we drop `stats` as well? Yes stats could theoretically be the body of a dataset, but that feels a little shoehorned to me.

Feels like a nice grouping to have `render`, `sql`, and starlark transformations in one spot.

PR #49, opened 2020-04-17
RFC: replace publish/add with push/pull

b5:
First pass at a big overhaul to the way we move versions around.

### DIFF
--- a/text/XXXX-replace_publish_clone_with_push_pull.md
+++ b/text/XXXX-replace_publish_clone_with_push_pull.md
@@ -1,0 +1,160 @@
+- Feature Name: Dataset Push & Pull
+- Start Date: <!-- (fill me in with today's date, YYYY-MM-DD) -->
+- RFC PR: <!-- (leave this empty) -->
+- Issue: <!-- (leave this empty) -->
+
+# Summary
+[summary]: #summary
+
+Replace `publish/add` with `push/pull`, which default to pushing all data, pulling only the latest version. A new `clone` command combines `pull` with `checkout`.
+
+# Motivation
+[motivation]: #motivation
+
+This RFC addresses two problems with the way we move versions around in qri:
+1. datasets published to the registry are creating weird states where users can't get versions
+2. the terms we use for moving versions around are confusing.
+
+The first problem comes down to improper defaults. Publishing should default to pushing _as much data as the remote is willing to accept_. Publishers sending data to a willing host should be trying to keep the network as robust as possible by making lots of copies.
+
+_Consumers_ of data on the other hand want just enough data to make use of the dataset, so the present default of only grabbing the latest version works _as a start_. Consumers are having problems satisfying deeper use cases that require prior versions because we don't present an easy API for common steps beyond grabbing the latest version of a dataset. It's currently too hard to:
+* fetch all versions of a dataset
+* fetch a specific historical version of a dataset
+* re-fetch once a dataset update has been published
+
+
+The second problem comes from poor metaphorical choices. Moving dataset histories around with "publish" and "add" is causing too much confusion. Making things worse, we're still stuck in a part-way transition to renaming "add" to "clone". What's worse, our concept of "clone" doesn't preently match
+
+We should remove `add` and `publish`, replacing them with three new commands:
+* push: upload datasets to qri peers
+* pull: download datasets from the qri network
+* clone: pull + checkout
+
+I like push & pull because they're clear metaphors based on physical verbs. push & pull both describe who is initiating the action (you) and the directionality of the action. In this world clone becomes a porcelain command that properly matches the way git behaves.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## New Commands
+I'm going to focus on writing help text that explains the proposed commands. The help text should describe the commands well enough to understand what this RFC is proposing.
+
+
+### push: upload datasets to qri peers
+```
+$ qri push --help
+Push sends datasets to a remote qri node. A push updates the dataset log on the
+remote and sends as many versions as the remote is willing to accept. A remote
+peer determines it's own rules on which versions to accept, but by default
+favour the latest versions of a dataset.
+
+If no remote is specified, qri pushes to the registry. But pushes 
+can just as easily send to another peer that is configured to accept datasets 
+from you. For details on  setting peers for dataset exchange read this doc:
+https://qri.io/docs/hosting-datasets
+
+Usage:
+  qri push [REMOTE] DATASET [DATASET] [flags]
+
+Aliases:
+  publish
+
+Examples:
+  # push a dataset to the registry
+  $ qri push me/dataset
+
+  # push a specific version of a dataset to the registry:
+  $ qri push me/dataset@/ipfs/QmHashOfVersion
+
+  # push only the latest two versions to a peer named xiao
+  $ qri push xiao me/dataset --revisions 2
+
+Flags:
+  -l, --logs               send only dataset history
+  -r, --revisions          send only the lastest commit data
+```
+
+### pull: download datasets from the qri network
+```
+$ qri pull --help
+Pull downloads datasets and stores them locally, fetching the dataset log and
+dataset version(s). By default pull only fetches the latest version of a dataset
+
+By default pull fetches 
+
+Because pull grabs the latest version by default, it's common to 
+
+If qri connect is running
+
+Usage:
+  qri pull [REMOTE] [DATASET] [flags]
+
+Examples:
+  # download a dataset log and latest version
+  $ qri pull b5/world_bank_population
+
+  # download a dataset log and all versions
+  $ qri pull b5/world_bank_population --all
+
+  # pull a dataset from 
+  $ qri pull ramfox b5/world_bank_population
+
+Flags:
+  -l, --logs              save only dataset history
+  -r, --revisions         fetch a set number of versions from the latest commit
+  -a, --all               fetch all versions, overrides revisions
+  -k, --keep-files        when in a working directory, don't update files
+
+```
+
+`revisions` and `all` flags mimic the way `qri remove` works.
+
+### clone: pull and checkout a dataset
+
+We already have "clone" language in qri cloud, and have intended to rename `add` to `clone` for some time. The problem is add isn't the same thing as clone when compared to git. Removing the idea of publish & unpublish gives us back a little cognitive overhead that we can use for the classic combo of "get the latest version and checkout a directory
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+* history syncing 
+* dsync needs support for a manifest with multiple roots, multiple pins
+* receivers may not honor all block push or pin requests
+* pushing with `--latest` makes push behave like pull
+* pulling with `--all` makes pull behave like default push
+* All Peers are Potential Remotes
+* IPFS is the natural fallback network
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+### Push & Pull don't work like git
+This is a problem, but I think the deviation is justified because we want different things from dataset version control than code version control. Datasets can get big. Versions of datasets can get bigger.
+
+We need to make it _very_ clear in documentation that pulling a dataset doesn't get you all versions by default.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+### Make pull fetch all versions when the total size is below a certain threshold
+
+# Prior art
+[prior-art]: #prior-art
+
+<!-- Discuss prior art, both the good and the bad, in relation to this proposal.
+A few examples of what this can include are:
+
+- Does this feature exist in other places and what experience have their community had?
+- For community proposals: Is this done by some other community and what were their experiences with it?
+- For other teams: What lessons can we learn from what other communities have done here?
+- Papers: Are there any published papers or great posts that discuss this? If you have some relevant papers to refer to, this can serve as a more detailed theoretical background.
+
+This section is intended to encourage you as an author to think about the lessons from other projects, provide readers of your RFC with a fuller picture.
+If there is no prior art, that is fine - your ideas are interesting to us whether they are brand new or if it is an adaptation from other languages.
+
+Note that while precedent set by other projects is some motivation, it does not on its own motivate an RFC.
+Please also take into consideration that Qri sometimes intentionally diverges from other projects. -->
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+Should pull use the list of remote sources
--- a/text/XXXX-replace_publish_clone_with_push_pull.md
+++ b/text/XXXX-replace_publish_clone_with_push_pull.md
@@ -1,12 +1,12 @@
 - Feature Name: Dataset Push & Pull
-- Start Date: <!-- (fill me in with today's date, YYYY-MM-DD) -->
-- RFC PR: <!-- (leave this empty) -->
+- Start Date: 17-04-2020
+- RFC PR: [#49](https://github.com/qri-io/rfcs/pull/49)
 - Issue: <!-- (leave this empty) -->
 
 # Summary
 [summary]: #summary
 
-Replace `publish/add` with `push/pull`, which default to pushing as much data as possible, and pulling only the latest version. Implient _automatic following_ within `checkout` to get git-clone like behaviour. Introduce _retention strategies_, a subject for later RFCs.
+Replace `publish/add` with `push/pull`, which default to pushing as much data as possible, and pulling only the latest version. Implement _automatic pulling_ within `checkout` to get git-clone style behaviour. Introduce _retention strategies_, a subject for later RFCs.
 
 # Motivation
 [motivation]: #motivation
@@ -15,7 +15,7 @@ This RFC addresses two problems with the way we move versions around in qri:
 1. datasets published to the registry are creating weird states where users can't get versions
 2. the terms we use for moving versions around are confusing.
 
-The first problem comes down to improper defaults. Publish currently sends _only the latest version_ of a dataset. If a user creates five versions, then runs publish, only the fifth version is pushed, leaving any party intersted in earlier versions with only the author as a viable source of versioned data. If the author isn't online and dialable, those prior versions are unavailable to the network. On the upside, our work on logbook and log syncronization means that at least the _history_ of the dataset is visible.
+The first problem comes down to improper defaults. Publish currently sends _only the latest version_ of a dataset. If a user creates five versions, then runs publish, only the fifth version is pushed, leaving any party intersted in earlier versions with only the author as a viable source of versioned data. If the author isn't online and dialable, those prior versions are unavailable to the network. On the upside, our work on logbook and log synchronization means that at least the _history_ of the dataset is visible.
 
 Publishing should default to pushing _as much data as the remote is willing to accept_. Publishers sending data to a willing host should be trying to keep the network as robust as possible by making lots of copies.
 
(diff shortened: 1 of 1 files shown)


text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,160 @@
...
+  -a, --all               fetch all versions, overrides revisions
+  -k, --keep-files        when in a working directory, don't update files
+
+```
+
+`revisions` and `all` flags mimic the way `qri remove` works.
+
+### clone: pull and checkout a dataset
+
+We already have "clone" language in qri cloud, and have intended to rename `add` to `clone` for some time. The problem is add isn't the same thing as clone when compared to git. Removing the idea of publish & unpublish gives us back a little cognitive overhead that we can use for the classic combo of "get the latest version and checkout a directory
```
Mr0grog, line 113: This made me realize I didn’t see anything in the above about unpublishing. I’m assuming it’ll be part of `qri push`, but don’t feel clear on how it would work.

It doesn’t *feel* like there’s an obvious analogue to Git, where the full command is more like “push this local pointer to a tree of hashes to this remote pointer to a tree of hashes” — and you unpublish by pushing nothing to a previously existing branch pointer. (And, arguably, this is among the more confusing parts of Git that people use regularly.)

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,160 @@
...
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+* history syncing 
+* dsync needs support for a manifest with multiple roots, multiple pins
+* receivers may not honor all block push or pin requests
+* pushing with `--latest` makes push behave like pull
```
Mr0grog, line 122: Is this missing from the example documentation above? I had to look back and I don’t see it mentioned anywhere else.

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,160 @@
...
+* All Peers are Potential Remotes
+* IPFS is the natural fallback network
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+### Push & Pull don't work like git
+This is a problem, but I think the deviation is justified because we want different things from dataset version control than code version control. Datasets can get big. Versions of datasets can get bigger.
+
+We need to make it _very_ clear in documentation that pulling a dataset doesn't get you all versions by default.
```
Mr0grog, line 133: One of the things I was thinking about as I read this was: I’m not *usually* concerned about the cost of pulling a Git repo☨, but I often would be when it comes to data. It’s just more likely to get bigger faster, and I’m *probably* less likely to need to investigate or validate the history. Do you think there might be a feasible way to estimate the cost of pulling N revisions before actually doing so? (It doesn’t seem easy since Qri versions are trees with lots of common nodes between them rather than diffs that build up.)


*☨: Obviously there are exceptions for some bigger projects, but those situations are rare for most people and usually made as reasonable as possible by maintainers splitting up repos or squashing histories every so often. There aren’t similar history manipulation tools in Qri, and I’d bet many potential publishers or registries would be philosophically against squashing like that (that kind of thing feels somehow more sacred with *data* than a work product or tool like code).*

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,160 @@
...
+# Summary
+[summary]: #summary
+
+Replace `publish/add` with `push/pull`, which default to pushing all data, pulling only the latest version. A new `clone` command combines `pull` with `checkout`.
+
+# Motivation
+[motivation]: #motivation
+
+This RFC addresses two problems with the way we move versions around in qri:
+1. datasets published to the registry are creating weird states where users can't get versions
```
Mr0grog, line 15: I might just be getting lost because the solutions for this are mixed in with renaming things here, but after reading the rest of this doc, I don’t feel totally clear on exactly what the major issues with publishing/pushing are (or what kinds of weird states are getting created) or exactly how this proposal addresses them. I know I’ve run into at least one of these, but I don’t know what broader scope of issues looks like. (The changes described here still definitely seem like good improvements and clarifications, though!)

chriswhong: I also think we are on the right track.  I've found it hard to understand what exactly was published when I publish, or what I got when I used "add".  I'd also suggest the CLI and Desktop reinforce what is happening.

```
> qri push me/some_dataset
pushing 28 versions of chriswhong/some_dataset...
X of XX blocks pushed for version 12 August 2019 (Qmss4h552h3j2h33)...

> qri pull me/some_dataset
pulling latest version of chriswhong/some_dataset | 13 August 2019 (Qmfie4kjn2di3223)...
There are 17 prior versions. Use --all get all, or use the full dataset reference to get a specific version
```

text/XXXX-replace_publish_clone_with_push_pull.md
chriswhong: +1, we need clarification on unpublishing.  If I pushed directly to a peer, presumably I can't "take it back", but I *can* prevent that peer from seeing updates to the dataset log or pulling new versions.

If I push to cloud, does cloud check my peer for new updates, or just think it has the latest until I push again?  If I push to a peer, can the peer check for updates on its own or do I have to push again?

text/XXXX-replace_publish_clone_with_push_pull.md
b5: Super helpful early feedback. I wrote the solutions before the motivation section (problem statement). I'll add some concrete examples

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,160 @@
...
+$ qri push --help
+Push sends datasets to a remote qri node. A push updates the dataset log on the
+remote and sends as many versions as the remote is willing to accept. A remote
+peer determines it's own rules on which versions to accept, but by default
+favour the latest versions of a dataset.
+
+If no remote is specified, qri pushes to the registry. But pushes 
+can just as easily send to another peer that is configured to accept datasets 
+from you. For details on  setting peers for dataset exchange read this doc:
+https://qri.io/docs/hosting-datasets
```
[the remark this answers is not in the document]
b5: unwritten doc. in my head that URL is a landing page for all use cases outside of the standard push-to-registry pattern. Based on our understanding of who's using qri three primary activities will drive people there:

* datasets that are too big for the registry's limits (an error message should lead to the same URL when pushing a 300Mb+ dataset version)
* decentralized LOCKSS (lot of copies keep stuff safe)
* unlisted datasets - people who aren't working with private data _per se_, but don't want to announce their work to the world, and would instead prefer to keep their own backups / coordination points

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,160 @@
...
+
+Usage:
+  qri push [REMOTE] DATASET [DATASET] [flags]
```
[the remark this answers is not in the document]
b5: yep yep yep

b5:
Ok we  [talked about this a bunch on our community call](https://youtu.be/7ytQhmF3FXk?t=567)

Here are some of my notes from it.
* **Rico:** "the phrase publish let a user know that when they hit `publish`, it'll make things visible to the world"

* **Asmir:** coming to qri fresh, there's a lot of term confusion in general. I'd like to see things funneled down to fewer commands. Many people will want to use a happy path of one or two commands, and get away from 
* **Dustin:** agree with Asmir. We have an opportunity to reduce the complication with git. Having fewer commands with better defaults is better. For example, can clone be a flag on another command?
* **Rob:** We need to show this to real users & get their feedback
* **Chris:** I agree with Rob's comments on the RFC about removing datasets. I'm confused about how to unpublish.
* **Kasey:** I'm also curious about how to push or pull specific versions? 
* **b5:** you can pull specific versions with `qri pull use/dataset@/ipfs/VersionHash`. It'd also be nice to be able to reference a specific index from the log by number like "qri pull the 3rd commit"
* **dustmop:** what if we merged "clone" into "checkout"? We've struggled with weather or not to call out network / not network as a clear distinction. Does it matter that there's a network call?
* rob: there should be a flag like `--offline`
* b5: pushing by default should accept newest versions, drop from oldest
* asmir: You're potentially opening a huge can of worms here. What if I want to be an archive node that prefers OLD versions? You're also biasing the network towards recency, making it probable that the network will forget original versions, making the whole audit trail at the beginning drop. This needs to be worked out as an RFC of it's own.
* rob: this _absolutely_ needs to be configurable. What if instead as a default for now: "if it doesn't fit, it doesn't go".
* rusty: if you can know in advance of the push 

There's lots of great feedback here I'm going to work directly into the RFC

b5:
Look, reference section for this isn't written, but the rest of this is ready for review

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+# Motivation
+[motivation]: #motivation
+
+This RFC addresses two problems with the way we move versions around in qri:
+1. datasets published to the registry are creating weird states where users can't get versions
+2. the terms we use for moving versions around are confusing.
+
+The first problem comes down to improper defaults. Publish currently sends _only the latest version_ of a dataset. If a user creates five versions, then runs publish, only the fifth version is pushed, leaving any party intersted in earlier versions with only the author as a viable source of versioned data. If the author isn't online and dialable, those prior versions are unavailable to the network. On the upside, our work on logbook and log syncronization means that at least the _history_ of the dataset is visible.
+
+Publishing should default to pushing _as much data as the remote is willing to accept_. Publishers sending data to a willing host should be trying to keep the network as robust as possible by making lots of copies.
```
Arqu, line 20: I think we should elaborate on this a bit more. One of my concerns here is if you cannot fit the last X revisions. Do you send over the oldest Y versions or the newest Z. Also think this should potentially not be configurable on the user side and we should stick to 1 default otherwise it would end up just being erratic behavior for end users.

The resolution of this issue should be done on the recieving end ideally a "registry" supporting some sort of "version distribution" among peers to accomodate and spread out the data while for a "repo" it should be fine to outright reject requests based on its policy.

text/XXXX-replace_publish_clone_with_push_pull.md
Arqu: I like this a lot. It also fits nicely in the story of the SQL query prompt being able to show the size of data that will be fetched/scanned so that the query can be executed. Also think this would be useful for adjusting all sorts of behaviors, or at least have a simple guard for users like "hey you're going to download 1TB of data, are you sure (y/N)?"

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+  qri push [REMOTE] DATASET [DATASET...] [flags]
+
+Aliases:
+  publish
+
+Examples:
+  # push a dataset to the registry
+  $ qri push me/dataset
+
+  # push a specific version of a dataset to the registry:
```
ramfox, line 67: assuming this means it only pushes the specific version of the dataset to the registry. If so, adding an 'only' in there will clarify even more:
"push only this specific version of a dataset to the registry:"

Q:
Can you use -- revisions' in conjunction with a version? qri push me/dataset@/ipfs/QmHashOfVersion --revisions 2would push versionQmHashOfVersion` as well as the immediately preceding version

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+  -r, --revisions          send only the lastest commit data
+```
+
+### pull: download datasets from the qri network
+```
+$ qri pull --help
+Pull downloads datasets and stores them locally, fetching the dataset log and
+dataset version(s). By default pull only fetches the latest version of a dataset
+
+Because pull grabs the latest version by default, it's common to
```
ramfox, line 84: unfinished thought

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+### checkout can pull
+
+We already have a checkout command, this RFC proposes that checkout will do the default of `qri pull` if a user tries to checkout a dataset they don't have.
+
+```
+$ qri checkout --help
+Create a linked directory and write dataset files to that directory. 
+
+If the dataset you are checking out isn't in the qri database, the latest 
+version of the dataset will be pulled from the network first. To prevent network
```
ramfox, line 120: assuming this also works for a specified hash, if one is given, not just the most recent.

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+      --log-all            log all activity
+      --no-color           disable colorized output
+      --no-prompt          disable all interactive prompts
+      --repo string        provide a path to load qri from (default "/Users/b5/.qri")
+```
+
+### No more clone
+We already have "clone" language in qri cloud, and the original plan has been to rename `add` to `clone`. In discussion we've come to realize The problem is add isn't the same thing as clone when compared to git. Git needs `clone` because git does not have a built-in naming system. Clone copies a git repo _from a place_, usually a URL. 
+
+Because qri can resolve dataset names without additional information from the user we can add `pull` directly to `checkout`, and drop the term `clone` from our lexicon. We can use this saved term later for something like "following" a dataset. See the unsreolved questions section for more info.
```
ramfox, line 145: spelling: "unsreolved" => "unresolved"

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+
+### Automatic Pulling & Networking UX
+Giving `checkout` the power to pull `pull` when data isn't local is the first concrete example of _automatic pulling_. This change means checkout by default assumes it can use the network to fulfill requests, which is is a shift in our user experience. We've wrestled with weather qri should default to making network requests for users for some time. As an example `qri --help` has a set of "Network commands", implying the other commands in qri _don't_ use the network. This RFC proposes we break that distinction, and assume that any command within qri _might_ use the network, unless instructed otherwise with either a configuration value or an `--offline` flag.
```
ramfox, line 148: spelling: "weather" => "whether"

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+
+Automatic pulling requires us to make a decision on this, and I think it's the right choice is to assume network access by default. We can take a cue from package managers, which also have a namespace, and assume network acess, automatically making network requests when the user seeks datasets they don't have. `npm install some_package` downloads from the web. Both npm and qri have name resolution systems that make this process possible. From a UX perspective npm lets me _ignore_ the network and focus on writing software.
```
ramfox, line 150: sp: "network acess" => "network access"

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+
+We've struggled with the word "network" because unlike npm, qri also has a peer-2-peer system, but I think the simple answer to needing assets that may only be available in a p2p context are to show an error asking users to run `qri connect`.
```
ramfox, line 152: IIInnnnnteresting.
Haven't dug into this in a while, but are we anticipating most remotes will be accessed via HTTP? Or would pushing to a remote (a trussed up peer) be a p2p interaction? Both, depending on the multiaddress/uri that is given to talk to the remote?

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+
+Defining possible retention strategies is beyond the scope of this RFC, but we can define the term here, and some of the problems retention strategies both create and address. It's also worth noting that retention strategies apply to _both_ pulling and pushing.
+
+In discussions we've come to realize that retention strategies are a major concern for the health of the qri network as a whole. For example: if all peers default to rolling-window strategy, we end up with a network of peers that "forget" early versions when the max size is exceeded. This destroys the auditability of a dataset, because no one retains the history. 
+
+A _retention_ strategy is also a _replication_ strategy. Put another way: which versions peers choose to keep determines the versions availabile to the network.
+
+### Remotes default to a `none` retention strategy
+If the user requests to push N versions for Y data size, the typical response is to accept all pushed verions, so long as the total disk space consumed is less than `remote.acceptSizeMax`. When a user tries to push a number of versions above the max,  `__ Mb` over the bar
```
ramfox, line 179: I believe this is an unfinished thought. What happens when a user tries to push a number of versions above the max? Error response and no pushing actually occurs?

ramfox:
V cool. I think this is definitely the right way to go.

There are two major things I think need to be discussed.

The first is the language and logic around what makes a dataset 'published' how we calculate that. Is the word/concept of 'published' being removed from our lexicon? Can datasets be considered 'published' or are only versions 'published'. How do we know or determine if a dataset is 'pushed' to more then one place? Do we need to keep track of it? Mostly thinking ahead to how we show/display this on the desktop and what information we will need to convey the correct state of a dataset.

The other major thing is 'unpublish'/remove from a remote, but I believe some of that is talked about in the storage rfc

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,160 @@
+- Feature Name: Dataset Push & Pull
+- Start Date: <!-- (fill me in with today's date, YYYY-MM-DD) -->
+- RFC PR: <!-- (leave this empty) -->
+- Issue: <!-- (leave this empty) -->
+
+# Summary
+[summary]: #summary
+
+Replace `publish/add` with `push/pull`, which default to pushing all data, pulling only the latest version. A new `clone` command combines `pull` with `checkout`.
```
dustmop: Discussed in a recent meeting, but just repeating here: I think we can avoid adding a `clone` command by just letting `checkout` just pull the data first if the user is connected.

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
+- Feature Name: Dataset Push & Pull
+- Start Date: <!-- (fill me in with today's date, YYYY-MM-DD) -->
+- RFC PR: <!-- (leave this empty) -->
+- Issue: <!-- (leave this empty) -->
+
+# Summary
+[summary]: #summary
+
+Replace `publish/add` with `push/pull`, which default to pushing as much data as possible, and pulling only the latest version. Implient _automatic following_ within `checkout` to get git-clone like behaviour. Introduce _retention strategies_, a subject for later RFCs.
```
feep, line 9: s/Implient/Implement/ ?

text/XXXX-replace_publish_clone_with_push_pull.md
b5: It'd be fun to be able to combine with revisions, definitely a _very_ advanced usage. I don't see why we couldn't support it. Might with writing test cases for the `revisions` calculation function

b5:
Thanks to @ramfox's feedback:

> The first is the language and logic around what makes a dataset 'published' how we calculate that.

This is now covered in the "access" RFC.


The other major thing is 'unpublish'/remove from a remote, but I believe some of that is talked about in the storage rfc

This is now covered in the "delete/drop" RFC.

text/XXXX-replace_publish_clone_with_push_pull.md
b5: point taken on the unpublishing bit missing. I'ver written another rfc that proposes `qri drop` for this: https://github.com/qri-io/rfcs/pull/53

text/XXXX-replace_publish_clone_with_push_pull.md
b5: oops, I'll add!

text/XXXX-replace_publish_clone_with_push_pull.md
b5: into it. I'm going to add it the the "unresolved questions" section.

text/XXXX-replace_publish_clone_with_push_pull.md
b5: _kiiiiinda_. We'll need to land and rely on dscache before we can get there. Currently you can only checkout HEAD

feep:
Links:

> "[access](https://github.com/qri-io/rfcs/pull/54)" RFC.
> "[delete/drop](https://github.com/qri-io/rfcs/pull/53)" RFC.

text/XXXX-replace_publish_clone_with_push_pull.md
b5: yep, we currenty have implementations for _both_ HTTP and P2P protocols for `dsync` and `logsync`. Right now remotes can be either HTTP addresses or multiaddrs

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,260 @@
...
+  $ qri push me/dataset
+
+  # push a specific version of a dataset to the registry:
+  $ qri push me/dataset@/ipfs/QmHashOfVersion
+
+  # push only the latest two versions to a peer named xiao
+  $ qri push xiao me/dataset --revisions 2
+
+Flags:
+      --latest             send only histry and the most recent dataset version
```
ramfox, line 71: sp "histry" => "history"

b5:
Ok, it's in! Thanks for everyone's input!

PR #50, opened 2020-04-22
RFC: storage command

b5:
RFC to support push/pull with storage manipulation commands. Early comments welcome!


text/XXXX-storage_command.md
```diff
@@ -0,0 +1,178 @@
...
+# Motivation
+[motivation]: #motivation
+
+We're missing a command for managing storage. Work on _retention strategies_ from another RFC has forced the issue. We need a robust command for managing storage locally, and a way to manually reach out to a remote to make storage management asks of them.
+
+Remove is currently covering _some_ of this territory for us, but not nearly enough. Adding storage management to `remove` would be overloading `remove`'s primary purpose of manipulating history. So we need a new command!
+
+This RFC proposes starting with two subcommands under a new command: `qri storage`:
+* `qri storage stat`
+* `qri storage drop`
```
Arqu, line 20: This might be more a personal feel type of thing, but drop sounds like a very deterministic action and feels like it should also be targeted at something. I propose for something that is more a "should" rather than "must" for this operation as ti is more opportunistic. My example would be "clean" where it's expected that a user has to do a confirmation step and acknowledge you will clean out X datasets/versions and free up Y space. The level of verbosity can be adjustable.

Another concern is how do we recover if a user manually pokes around their file system and sync up state (which is a concern regardless of this new command)

text/XXXX-storage_command.md
```diff
@@ -0,0 +1,178 @@
...
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+<!-- - Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this? -->
+
+# Prior art
+[prior-art]: #prior-art
+
+
```
Arqu, line 166: AFAIK Docker has a similar "process" to clean up old/unused/build images from your machine. The ramifications are not really comparable but offer some parallel. One of the things they do is clean out "inactive/stopped" images which might make sense that we start from some sort of least used datasets when thinking about purging and maybe having different levels of how aggresive the purge is.

text/XXXX-storage_command.md
```diff
@@ -0,0 +1,178 @@
...
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+## Knowing which datasets depend on a block
+One of the problems this RFC doesn't tackle properly is _knowing_ if removing a block will free up storage space, because we aren't leveraging pin reference counting. In an ideal world we'd have a low-level command like this:
+
+```
+qri storage deps: for a given block, show which dataset versions depend on that that block
```
Arqu, line 175: Do we also have, or at least want to have, de-duplication across multiple datasets based on these blocks. In that case reference counting is a must (and opens us up to a whole slew of weird race conditions with peers unless we guard with some sort of replication strategy?)

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+  qri push [REMOTE] DATASET [DATASET...] [flags]
+
+Aliases:
+  publish
+
+Examples:
+  # push a dataset to the registry
+  $ qri push me/dataset
+
+  # push a specific version of a dataset to the registry:
```
ramfox, line 67: assuming this means it *only* pushes the specific version of the dataset to the registry. If so, adding an 'only' in there will clarify even more:
"push only this specific version of a dataset to the registry:"

Q:
Can you use `-- revisions' in conjunction with a version? 
`qri push me/dataset@/ipfs/QmHashOfVersion --revisions 2`
would push version `QmHashOfVersion` as well as the immediately preceding version

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+  -r, --revisions          send only the lastest commit data
+```
+
+### pull: download datasets from the qri network
+```
+$ qri pull --help
+Pull downloads datasets and stores them locally, fetching the dataset log and
+dataset version(s). By default pull only fetches the latest version of a dataset
+
+Because pull grabs the latest version by default, it's common to
```
ramfox, line 84: unfinished

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+      --log-all            log all activity
+      --no-color           disable colorized output
+      --no-prompt          disable all interactive prompts
+      --repo string        provide a path to load qri from (default "/Users/b5/.qri")
+```
+
+### No more clone
+We already have "clone" language in qri cloud, and the original plan has been to rename `add` to `clone`. In discussion we've come to realize The problem is add isn't the same thing as clone when compared to git. Git needs `clone` because git does not have a built-in naming system. Clone copies a git repo _from a place_, usually a URL. 
+
+Because qri can resolve dataset names without additional information from the user we can add `pull` directly to `checkout`, and drop the term `clone` from our lexicon. We can use this saved term later for something like "following" a dataset. See the unsreolved questions section for more info.
```
ramfox, line 145: "unsreolved" => "unresolved"

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+
+### Automatic Pulling & Networking UX
+Giving `checkout` the power to pull `pull` when data isn't local is the first concrete example of _automatic pulling_. This change means checkout by default assumes it can use the network to fulfill requests, which is is a shift in our user experience. We've wrestled with weather qri should default to making network requests for users for some time. As an example `qri --help` has a set of "Network commands", implying the other commands in qri _don't_ use the network. This RFC proposes we break that distinction, and assume that any command within qri _might_ use the network, unless instructed otherwise with either a configuration value or an `--offline` flag.
```
ramfox, line 148: "weather" => "whether"

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+
+Automatic pulling requires us to make a decision on this, and I think it's the right choice is to assume network access by default. We can take a cue from package managers, which also have a namespace, and assume network acess, automatically making network requests when the user seeks datasets they don't have. `npm install some_package` downloads from the web. Both npm and qri have name resolution systems that make this process possible. From a UX perspective npm lets me _ignore_ the network and focus on writing software.
```
ramfox, line 150: "network acess" => "network access"

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+
+We've struggled with the word "network" because unlike npm, qri also has a peer-2-peer system, but I think the simple answer to needing assets that may only be available in a p2p context are to show an error asking users to run `qri connect`.
```
ramfox, line 152: IIInnnnnteresting.
Haven't dug into this in a while, but are we anticipating most remotes will be accessed via HTTP? Or would pushing to a remote (a trussed up peer) be a p2p interaction? Both, depending on the multiaddress/uri that is given to talk to the remote?

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+
+Defining possible retention strategies is beyond the scope of this RFC, but we can define the term here, and some of the problems retention strategies both create and address. It's also worth noting that retention strategies apply to _both_ pulling and pushing.
+
+In discussions we've come to realize that retention strategies are a major concern for the health of the qri network as a whole. For example: if all peers default to rolling-window strategy, we end up with a network of peers that "forget" early versions when the max size is exceeded. This destroys the auditability of a dataset, because no one retains the history. 
+
+A _retention_ strategy is also a _replication_ strategy. Put another way: which versions peers choose to keep determines the versions availabile to the network.
+
+### Remotes default to a `none` retention strategy
+If the user requests to push N versions for Y data size, the typical response is to accept all pushed verions, so long as the total disk space consumed is less than `remote.acceptSizeMax`. When a user tries to push a number of versions above the max,  `__ Mb` over the bar
```
ramfox, line 179: unfinished, I believe. What happens when a user tries to push a number of versions above the max? Error response and no pushing actually occurs?

text/XXXX-storage_command.md
```diff
@@ -0,0 +1,178 @@
...
+### purge: remove stored datasets without affecting history
+```
+$ qri storage purge --help
+Purge deletes dataset version data, freeing up disk space. Purge does not edit 
+history. To edit version history, use the remove command.
+
+Dataset versions take up disk space, and at some point need to be managed.
+Becase datasets and storage capacity vary dramatically the need for this
+
+You can also use purge to request a remote drop version data. This is often
```
ramfox, line 90: great this answers a bunch of my questions.

ramfox: except that the thought isn't finished here, lol

text/XXXX-storage_command.md
ramfox: b/c of ipfs we have natural dedupe on identical blocks added to the ipfs store from any datasets added. So v good point about clean up, need to ensure that we aren't depending on these blocks in other contexts before removing, although this feels like a problem ipfs would have had to tackle when 'unpinning' already.

text/XXXX-storage_command.md
```diff
@@ -0,0 +1,178 @@
...
+  $ qri storage stat
+
+  # show storage stats for a dataset
+  $ qri storage stat b5/world_bank_population
+
+  # show overlap between two different dataset histories
+  $ qri storage stat b5/world_bank_population b5/country_codes
+
+  # show storage cost of a specific commit:
+  $ qri storage stat
```
ramfox, line 57: unfinished

text/XXXX-storage_command.md
```diff
@@ -0,0 +1,178 @@
...
+  Links [][2]int `json:"links"` // links between nodes
+  Nodes []string `json:"nodes"` // list if CIDS contained in the DAG
+}
+```
+
+Manifests assume they describe one graph with a single root. To represent multiple graphs in the same manifest, we'll need to make one change:
+
+```go
+type Manifest struct {
+  Graphs [][][2]int `json:"links"` // links between nodes
```
ramfox, line 140: Does this have an impact on dsync?

Is one entire push process come off of one dag.Info, that iterates over the graphs in the manifest? Or is this just for storage checking purposes?

Also, where in the sync process does the 'rejection' happen based on the storage's retention strategy? If the entire dsync details arrives in the dag.Info (because we have the manifests of each version), then we can do it immediately.

text/XXXX-storage_command.md
b5: truth. the pin API maintains reference counts for each block under the hood. That's why the command to collect & delete 0 reference-count blocks is called `ipfs repo gc` 😄

text/XXXX-replace_publish_clone_with_push_pull.md
```diff
@@ -0,0 +1,258 @@
...
+When pushing a dataset, the receiver is in charge of what they will accept. A remote can reject a push request for any reason. One of the primary reasons for rejecting a push is do to size constraints. Qri has a configuration value `remote.acceptSizeMax` that currently polices the maximium size of a dataset version. Any dataset version that exeeds this value is rejected by default.
+
+This RFC proposes changing the `remote.acceptSizeMax` value in configuration to apply to the on-disk size of _the entire dataset history_. As an example: if `acceptSizeMax` is `300Mb` that quota can be filled with one version that is `300Mb` in size, three totally unique `100Mb` versions, or any other combination where the deduplicated size of all versions is below the maximum threshold. 
+
+We say totally unique versions because qri de-duplicates versioned data. A block of dataset data may be used in any number of versions, and should only count towards the size threshold once.
+
+### Retention Strategies
+This RFC introduces the term _retention strategy_ to describe the approach a peer takes to pruning versions when `acceptSizeMax` is exceeded. A few potential strategies, and their names might be:
```
Arqu, line 164: I'm still concerned with what the "default" should be here. If we have too many of any specific type it can negatively impact the network health.
- None: too many of these lead to an only latest N that fit are really available until storage is saturated at which point we have no new network updates (highly susceptible to `acceptSizeMax` and the specific dataset at hand)
- Archival:  too many of these lead to only the earliest N are available to the network (also think these are much more storage heavy but great for long term network health)
- Rolling Window: same as none only that we have just the latest N (I'm especially conerned with edge cases like if a dataset size is near the limit which means we only keep the latest copy)

We should have a strategy on how to balance this out. Think rolling window is the best default as it can be just left running and at least have the latest with archival nodes existing to balance things out. We should definitely investigate the best rations for network health and monitor that closely (metrics are important here). 

As I see it - users are typically `rolling-window`, remotes are `none` and registries are `archival`

text/XXXX-storage_command.md
chriswhong: +1, I end up using `docker system prune` a lot, it would be good to have some easily choosable options for clearing older versions of datasets, or datasets that have not been accessed recently, etc

text/XXXX-storage_command.md
b5: > My example would be "clean" where it's expected that a user has to do a confirmation step and acknowledge you will clean out X datasets/versions and free up Y space. The level of verbosity can be adjustable.

I love this, and feel like `clean` should be a command we add later that operates in _addition_ to something like `drop`, where `drop` is an explicit, manual command, and `clean` is an automated housekeeping command, much like [`brew cleanup`](https://discourse.brew.sh/t/why-does-brew-cleanup-work-so-well/1819). I'll add a note about automated cleanup to the "Unresolved Questions" section.

> Another concern is how do we recover if a user manually pokes around their file system and sync up state (which is a concern regardless of this new command)

We have to operate at all times with the assumption that the user might have done such a thing. in the IPFS case they can use the `ipfs` CLI to manually unpin and clean stuff. Qri needs to work in these cases, and (largely) does.

b5:
@Arqu, apologies, this PR erroneously includes a copy of the push/pull RFC, if I understand correctly, I think your comments about retention strategies apply to that RFC more than the storage one:

> I hate to say it but it might even warant a separate RFC for just the retention strategies and network health, split of from storage, accept policies/params and storage commands.

Totally agreed. Might not have made this clear enough, but I'm hoping these RFCs just set the stage for work on defining retention strategies, and the strategies themselves should be described in future RFCs. The one retention strategy we _do_ need right now is the `none` strategy, which when the storage max value is exceeded. In this "no strategy" case we need these `storage` commands to manually request a remote drop versions until we're under the limit again

text/XXXX-storage_command.md
b5: yes it most definitely does affect dsync. The reject can potentially happen in two places that  are part of the `dsync receive` lifecycle:

1. if the [dsync push preCheck hook](https://github.com/qri-io/dag/blob/master/dsync/dsync.go#L352) returns an error
2. if the [dsync finalCheck hook](https://github.com/qri-io/dag/blob/5cd938b03caf29397ad78d79b4f4830dd48ae54f/dsync/dsync.go#L415-L418) returns an error.

qri implements these dsync hooks to do storage checks and supply a higher-order set of hooks. Qri always checks `acceptSizeMax` in the `dsync.FinalCheck` hook.

> Also, where in the sync process does the 'rejection' happen based on the storage's retention strategy? If the entire dsync details arrives in the dag.Info (because we have the manifests of each version), then we can do it immediately.

Totally agreed. This is an optimization we should be shipping for dsync.

b5:
Ratified! Thanks all

PR #51, opened 2020-04-23
RFC: overhaul configuration

b5:
based on conversations with @Arqu & @dustmop, our config story needs to be better.


text/XXXX-config_revision.md
```diff
@@ -0,0 +1,268 @@
...
+All that said, using env var configuration should be _optional_, meaning a user employing qri as a library can provide an option to the `lib.NewInstance` constructor function that ignores all environment variables.
+
+### Config Hierarchy
+
+Configuration details should be unified into a single data structure loaded in a predefined hierarchy. From least that hierarchy should be:
+
+* default configuration
+* qri repo configuration file
+* global command-line flags
+
```
Arqu, line 61: I'd like to add that ENV vars should still remain in this hierarchy and that it's ok for an env var to specify a config and not just point to some. However it's context is and should be limited to the session itself and as such is pretty high up in the hierarchy as it's typically treates as a user supplied default without neccesarily materializing this change into the config.

As such I'd squeze it in between repo config and cmd flags.

text/XXXX-config_revision.md
```diff
@@ -0,0 +1,268 @@
...
+- type: ipfs
+  options:
+    path: ./ipfs
+    api: true
+    pubsub: false
+- type: local
+- type: http
+```
+
+### Expanding Store to Filesystems
```
Arqu, line 206: Not really sure I get this completely. Maybe we could dive deeper on a call?
I like the gist of it but seems like this is a _much_ deeper topic than this covers and would like to understand on where it carries us to in the context of the config RFC itself.

text/XXXX-config_revision.md
```diff
@@ -0,0 +1,268 @@
...
+the above filesystem example shows a `ipfs` filesystem type with a `config.path` field. This value should now the be canonical source of where to load.
+
+the global `ipfs-path` flag should also be removed:
+```
+      --ipfs-path string   override IPFS path location (default "/Users/b5/.ipfs")
+```
+
+### Relative Paths in configuration files are relative to the file
+If a configuration value specifies a filepath, and the given path is relative, that path should be relative to the location of the config file.
+
```
Arqu, line 237: Agree on this as it resolves ambiguity, however this should be properly documented behavior somewhere.

text/XXXX-config_revision.md
```diff
@@ -0,0 +1,268 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+### Per-field env vars
+We could support setting configuration on a per-field basis. I'm not sure that's totally necessary at this point.
```
Arqu, line 247: Think that might be overkill for our usage and would stick with using env vars for context specific tasks like:
- specifying another config
- specifying repo
- p2p and some other networking params

Would not like to have to support things like `no-color` as an env var unless we will invest the time and energy to have a fully dynamic ENV vars parser which we can just hot plug.

text/XXXX-config_revision.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Attach config to context.Context
+We could build a sophisticated system of attaching configuration to request contexts, to support scoping configuration on a per-request basis. We might need to in the future, but it's overkill for the moment.
```
Arqu, line 250: I know it's a thought for the future, however if we were to do this there's a lot of things to consider like _who's_ config takes presedence and most probablt needs to be dealt with on the field level. While ok when talking in local context, interacting with remotes makes this a whole lot more complicated.

text/XXXX-config_revision.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Config changes emit events
+We could also use the event bus to propagate changes to configuration. Again, overkill for now, but we could do so later.
+
```
Arqu, line 254: Not sure how necessary this is if we keep the config state centralized in the codebase. Though UI might be more sensitive to this for things like P2P on/off toggles or UI hints about the state.

text/XXXX-config_revision.md
```diff
@@ -0,0 +1,268 @@
...
+awesome-go has an entire section of configuration libraries https://awesome-go.com/#configuration. Some of the ones I found intersting:
+* hot-reloading config: https://github.com/lalamove/konfig
+* env var config with lots of bells& whistles: https://github.com/kelseyhightower/envconfig
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+Some configuration things are out of scope for this RFC, but should be discussed later:
+* Distinguish runtime _options_ from static _configuration_.
+* hot-swapping configuration.
```
Arqu, line 268: Think all operations relying on config should read from config directly for each time they run. Thing's that rely on config on startup should not be hot-swappable IMHO and should require an instance restart. Thus hot-swapping should be a given if we properly utilize the new config store.

text/XXXX-config_revision.md
```diff
@@ -0,0 +1,268 @@
...
+| `IPFS_SETUP_CONFIG_DATA` | JSON data when initializing IPFS      |
+| `QRI_PRIVATE_KEY`        | defined in the private key RFC        |
+| `QRI_BACKTRACE`          | show stack trace output on crash      |
+
+
+_note: I (b5) haven't had a chance to vet our dependencies for env vars they are sensitive to, so this list is only complete for the purpose of this RFC._
+
+This RFC proposes environment variables are _not_ a direct source of configuration. Instead environment variables either point to a path to load configuration from, or contain a payload to create an initial configuration file.
+
+`QRI_SETUP_CONFIG_DATA` as the only supported is only honored by `qri setup`. We need `SETUP` env variables for operating within containerized contexts, but that's about it.
```
ramfox, line 46: A bit confused about what the first sentence is saying here.

"`QRI_SETUP_CONFIG_DATA`  is only honored by `qri setup`." ??

text/XXXX-config_revision.md
```diff
@@ -0,0 +1,268 @@
...
+Render:
+  defaultTemplateHash: /ipfs/QmeqeRTf2Cvkqdx4xUdWi1nJB2TgCyxmemsL3H4f1eTBaw
+  templateUpdateAddress: /ipns/defaulttmpl.qri.io
+Stats: null
+Update:
+  address: ""
+  daemonize: false
+  type: fs
+```
+
```
ramfox, line 134: Assuming this next section of yaml is what we are keeping?
If so, needs an explanation
"The following fields should remain or be added:"

text/XXXX-config_revision.md
```diff
@@ -0,0 +1,268 @@
...
+  allowremoves: false
+  enabled: false
+  requireallblocks: false
+Remotes: null
+Repo:
+  type: fs
+Revision: 2
+Filesystems:
+- type: ipfs
+  options:
```
ramfox, line 198: This crosses over w/ the ipfs repo move question, but are we going to give a space for folks to config ipfs ports if they so choose (if they do not want to default to 0/random)? If so, should that live in this section? Or if someone wants to do that, should we assume they will know to go into the ipfs config to change? Wondering as we used to have to manual change this when we used multiple repos on the same machine for dev purposes, and defaulting to any available port would fix this conflict anyway, but wanted to bring up if it should be surfaced/able to be controlled via config.

text/XXXX-config_revision.md
```diff
@@ -0,0 +1,268 @@
...
+Filesystems:
+- type: ipfs
+  config:
+    path: ~/.ipfs
+    api: true
+    pubsub: false
+- type: local
+- type: http
+```
+
```
ramfox, line 220: Interesting. This brings me to another rfc (prfc) but assuming this is what we would rely on to know which systems to check for resolution in `ResolveRef`

text/XXXX-config_revision.md
```diff
@@ -0,0 +1,268 @@
...
+The first win: users could remove `local` or `http` to disable qri's access to the local filesystem or HTTP for the purpose of reading data. We can also expand for later filesystems & options. an `s3` implementation would be nice. this is also how we'd do `dat` integration.
+
+We could use a convention of the default write destination being the first listed filesystem, and present a big warning if users configure anything other than `ipfs` in the first position.
```
ramfox, line 223: Could be better to explicitly having a `write: true` flag in the filesystem's config and warn if no flag is specified or if it's not in specified under `ipfs`. Then it's easier for a user to know what's going on just by looking at the config.

text/XXXX-config_revision.md
ramfox: If desktop is running, this means that any time user switches a flag from local to network or (when we create a settings page) edits any settings that affect the backend, we would have to kill the qri process and then reconnect, or the user would have shut down the app and restart. Which... might be fine, especially the former option and when talking about general settings, but honestly toggling between `local` and `network` feels like something user will do a lot, casually, and we should make it seamless for them.

ramfox: If desktop is running, this means that any time user switches a flag from local to network or (when we create a settings page) edits any settings that affect the backend, we would have to kill the qri process and then reconnect, or the user would have shut down the app and restart. Which... might be fine, especially the former option and when talking about general settings, but honestly toggling between local and network feels like something user will do a lot, casually, and we should make it seamless for them.

text/XXXX-config_revision.md
b5: totally agreed. This is a big hot mess of work that I don't want to take on any time soon

text/XXXX-config_revision.md
b5: meaning this variable is only _used_ when running `qri setup`. All other times, it's value is ignored

text/XXXX-config_revision.md
b5: totally. having this in the alternatives section means we _won't_ do this. Sounds like we're all in agreement ok keeping ENV var configuration light.

text/XXXX-config_revision.md
b5: I think the first time we ship an on-off switch for p2p on desktop, it should be a fairly dramatic action that locks you out of desktop during the restart. This is one part UX, one part being conservative with network churn.

We want to keep people leaving and joining p2p networks lower if we can. Making it too easy to swap p2p will lead to people joining and leaving the network more than they need to. The added upside is an easier story around configuration. If anyone _really_ complains about p2p switching, we can re-evaluate.

text/XXXX-config_revision.md
b5: >  Or if someone wants to do that, should we assume they will know to go into the ipfs config to change?

We should assume they're going to manipulate the underlying IPFS `config` file in the repo itself. I don't want to "cross wires" between the IPFS config and qri config. All fields should be set in one or the other. The `options` fields you see here _can't_ be controlled via IPFS configuration.

text/XXXX-config_revision.md
b5: Agree, but I think we should save it for later. If multiple filestores set `write: true`, how do we know which destination to save it to? For now I think going with the convention of first-one-saves lets us delay dealing with this until later, and won't require deprecating a configuration field

b5: We can also just say for now: "qri only saves to IPFS". Then order wouldn't matter, and we can show an error if an IPFS filestore is missing

text/0029-config_revision.md
```diff
@@ -0,0 +1,291 @@
...
+  remote_name: https://remote_name_domain.com
+Repo:
+  type: fs
+Revision: 2
+```
+
+### The Filesystems field
+
+Qri is built atop a filesystem abstraction called [qfs](https://github.com/qri-io/qfs) that defines a common interface with numerous implementations. At runtime, qri composes a number of these filesystems together to provide access to HTTP assets, local files, and IPFS data. In qri v0.9.8, the `qfs.Filsystem` instance cannot be configured. 
+
```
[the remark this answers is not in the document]
b5: Great question. I've added a paragraph to the reference-level section describing how CAFS is on the way out.

we do spin up a `muxfs` now: https://github.com/qri-io/qri/blob/201bf7c0d34e792be61cc571fe2b1b0929735892/repo/buildrepo/build.go#L89-L102

Just need to make better use of that API 😄

text/0029-config_revision.md
```diff
@@ -0,0 +1,291 @@
...
+| `QRI_PATH`               | path to qri configuration data        |
+| `QRI_SETUP_CONFIG_DATA`  | JSON data to use when creating a repo |
+| `IPFS_SETUP_CONFIG_DATA` | JSON data when initializing IPFS      |
+| `QRI_PRIVATE_KEY`        | defined in the private key RFC        |
+| `QRI_BACKTRACE`          | show stack trace output on crash      |
+
+
+_note: I (b5) haven't had a chance to vet our dependencies for env vars they are sensitive to, so this list is only complete for the purpose of this RFC._
+
+This RFC proposes environment variables are _not_ a direct source of configuration. Instead environment variables either point to a path to load configuration from, or contain a payload to create an initial configuration file.
```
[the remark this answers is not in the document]
b5: Great sentence. I've added that last sentence into the RFC's reference-level section

b5:
Annnnnnnnnnd we're ratified! Thanks everyone for your input!

PR #52, opened 2020-04-30
RFC: externalize private key

b5:
I think we should land this one _before_ the configuration RFC. It informs design changes to our config, and takes some pressure off that RFC to explain how we're going to address https://github.com/qri-io/qri/issues/1306

### DIFF
--- a/text/XXXX-externalize_private_keys.md
+++ b/text/XXXX-externalize_private_keys.md
@@ -1,0 +1,111 @@
+- Feature Name: Externalize Private Keys
+- Start Date: <!-- (fill me in with today's date, YYYY-MM-DD) -->
+- RFC PR: <!-- (leave this empty) -->
+- Issue: <!-- (leave this empty) -->
+
+# Summary
+[summary]: #summary
+
+Separate Qri user private key from configuration, storing in the operating system's keyring, or superceeded by an envrionrment variable `QRI_PRIVATE_KEY`.
+
+# Motivation
+[motivation]: #motivation
+
+Qri users are getting themselves into un-recoverable states by manually deleting their `.qri` directory, which takes their private key with it, creating major problems when they try to re-add & change datasets. Because Qri is built on public-key infrastructure (PKI), deleting the private key has the effect of locking a user out of their own identity.
+
+More importantly, storing private keys in configuration is a security vulnerability. Coupling sensitive information with configuration renders the entire configuration sensitive.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+We inherited this bad habit of storing private keys in configuration as a "deal with it later" approach taken from IPFS, unlike Qri, IPFS identities are not "significant" in the sense that it's not tied to an identity intended for humans. Tying an identity to a private key means we need to deal with this problem much earlier.
+
+This RFC deprecates `profile.privKey` config value, moving the the private key to the user's operating system keyring. When no keyring is available, store the keyring in a seperate file called `QRI_PRIVATE_KEY` within the `.qri` directory, and present the user with a warning that they should copy this file to a safe location.
+
+In all cases the active private key can be overridden with an environment variable: `QRI_PRIVATE_KEY`. In all cases the key is expected to be a base64-encoded string.
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Doing the work of generalizing secret storage can be handled by an external package: https://github.com/zalando/go-keyring. go-keyring supports Windos, OS X, and Linux. The key is stored in a `(username,password)` combination, with the "username" being the user's ProfileID (hash of public key), and the password being the base64-encoded private key. OS Keyrings support namespacing, qri's namespace will be a reversed-domain-name constant:
+
+```go
+const qriKeyringServiceID = "io.qri.keys"
+```
+
+Once setup is complete, `config.Profile.ID` stores the ProfileID string this configuration is attached to. Starting a qri process uses this as the "username" in a `Get` call:
+
+```go
+privKey, _ := keyring.Get(qriKeyringServiceID, config.Profile.ID)
+```
+
+### Keys subsystem
+
+All of this work should happen in a new package within the qri codebase: `github.com/qri-io/qri/keys`. We need to takle key rotation next, followed by storing _many_ keys for doing access control. All of this work should happen in a new subsystem called `keys.Service`. `lib.Instance` gets a new field that points to this service.
+
+### Phasing out `config.Profile.PrivKey`
+This work should happen in phases:
+
+1. **copy to keyring, support env var**. In in the next release, add in the new APIs, prefer their use when present.
+2. **intiailize with an empty privKey value**. Once we know the keyring approach works across a number of different platforms in a wide release, switch default initialization behaviour to _not_ write the privKey value.
+3. **remove the field** Eventually, remove the field entirely with a repo migration.
+
+In all phases the privKey field should be the _lowest_ in the loading hierarchy. Keyring values will override `Profile.PrivKey`, and the env var will override all other values.
+
+### Recovering from a deleted `.qri` folder
+Once we've externalized the private keys, we can change the behaviour of `qri setup` to first scan the keyring under the `io.qri.keys` namespace for existing keys. If a key is found, we can ask the user if they'd like to create a new account, or recover an old one.
+
+We can bolster this with calls to the default registry for the discovered public keys to add username info to the list. Profile information may not exist on the default registry, so we'll need to default to showing just the raw ProfileID. We can support this process with docs.
+
+Instead of jumping straight to building the UI for switching, we can supply a `--recover` flag that lets a user supply a ProfileID to recover from:
+
+```
+$ qri setup --recover QmProfileID
+```
+
+listing available private keys in a keyring may also form the basis of profile switching in the future.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+### More OS-Specific Work.
+This increases our reliance on an OS-level API that sometimes doesn't exist, and runs the risk of deviating behaviour between platforms. 
+
+Fallback to a `QRI_PRIVATE_KEY` file whenever no keyring exists is necessary to keep things working when our keyring implementation fails, effectivly reverting us back to the setup we're in today. This time we at least have a separate file and a warning.
+
+### Upstream work to get recovery functionality
+The package I've suggested we build on has no high-level method for listing the keys in a namespace. We'e need to be able to do this, and haven't confirmed this will be possible across all platforms.
+
+### No Qri-specific security
+This RFC does not propose adding any key-specific security. If we used symmetric encryption on the private key itself, Qri could require a local password of it's own. In this case we're relying on the user's Operating System security architecture
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+A number of alternatives exist for externalizing secrets. We could store keys in the `.ssh` directory if it exists, and looking for guidance on 
+
+# Prior art
+[prior-art]: #prior-art
+
+* Keybase - We should really do a dive on the approach keybase uses for key management before moving on to any other keypair work. _Especially_ key rotation.
+* Textile wallet
+* Hashicorp Vault
+
+* https://github.com/keybase/go-keychain
+* https://github.com/zalando/go-keyring
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+### Loading order for configuration
+We haven't defined _when_ key loading happens in this process now that key loading is separate from configuration loading in the relation to seting up other services.
+
+### Profile Switching
+This RFC lays the groundwork for switching accounts _within_ qri, but this would have very strange results, as the data stored in `.qri` is mapped to a single user. We also don't have any UI for how this would work. 
+
+### Key Rotation
+We still haven't covered how to rotate this master private key in the event of a private key disclosure. That'll come later.
+
+### Multiple Keys 
+The "master key" is not the only key we need to store. In the future we'll end up needing to create all sorts of keys for managing access control. I'm wondering if we use something like [BIP0032](https://en.bitcoin.it/wiki/BIP_0032) for this.
--- a/text/0028-externalize_private_keys.md
+++ b/text/0028-externalize_private_keys.md
@@ -1,12 +1,12 @@
 - Feature Name: Externalize Private Keys
-- Start Date: <!-- (fill me in with today's date, YYYY-MM-DD) -->
-- RFC PR: <!-- (leave this empty) -->
+- Start Date: 2019-04-30
+- RFC PR: [#52](https://github.com/qri-io/rfcs/pull/52)
 - Issue: <!-- (leave this empty) -->
 
 # Summary
 [summary]: #summary
 
-Separate Qri user private key from configuration, storing in the operating system's keyring, or superceeded by an envrionrment variable `QRI_PRIVATE_KEY`.
+Separate Qri user private key from configuration, storing in the operating system's keyring, or superseded by an environment variable `QRI_PRIVATE_KEY`.
 
 # Motivation
 [motivation]: #motivation
@@ -20,7 +20,7 @@ More importantly, storing private keys in configuration is a security vulnerabil
 
 We inherited this bad habit of storing private keys in configuration as a "deal with it later" approach taken from IPFS, unlike Qri, IPFS identities are not "significant" in the sense that it's not tied to an identity intended for humans. Tying an identity to a private key means we need to deal with this problem much earlier.
 
-This RFC deprecates `profile.privKey` config value, moving the the private key to the user's operating system keyring. When no keyring is available, store the keyring in a seperate file called `QRI_PRIVATE_KEY` within the `.qri` directory, and present the user with a warning that they should copy this file to a safe location.
+This RFC deprecates `profile.privKey` config value, moving the the private key to the user's operating system keyring. When no keyring is available, store the keyring in a separate file called `QRI_PRIVATE_KEY` within the `.qri` directory, and present the user with a warning that they should copy this file to a safe location.
 
 In all cases the active private key can be overridden with an environment variable: `QRI_PRIVATE_KEY`. In all cases the key is expected to be a base64-encoded string.
 
@@ -72,7 +72,7 @@ listing available private keys in a keyring may also form the basis of profile s
 ### More OS-Specific Work.
 This increases our reliance on an OS-level API that sometimes doesn't exist, and runs the risk of deviating behaviour between platforms. 
 
-Fallback to a `QRI_PRIVATE_KEY` file whenever no keyring exists is necessary to keep things working when our keyring implementation fails, effectivly reverting us back to the setup we're in today. This time we at least have a separate file and a warning.
+Fallback to a `QRI_PRIVATE_KEY` file whenever no keyring exists is necessary to keep things working when our keyring implementation fails, effectively reverting us back to the setup we're in today. This time we at least have a separate file and a warning.
 
 ### Upstream work to get recovery functionality
 The package I've suggested we build on has no high-level method for listing the keys in a namespace. We'e need to be able to do this, and haven't confirmed this will be possible across all platforms.


text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
...
+$ qri setup --recover QmProfileID
+```
+
+listing available private keys in a keyring may also form the basis of profile switching in the future.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+### More OS-Specific Work.
+This increases our reliance on an OS-level API that sometimes doesn't exist, and runs the risk of deviating behaviour between platforms.
```
feep, line 73: I have been a _heavy_ macOS user over the years. I almost never use Keychain. If I used Keychain Access today, I’d have to poke around to find what I needed.

I know some novices who use it heavily (maybe because of Safari integration and sync), and I know super power users who wouldn’t know it any better than I do because they stick everything in 1Password or such.

Neither is key storage, I’m just talking about usage/familiarity.

OS native is good. But unless you need the user to look at the keys and interact with them, I don’t know if it gains anything from UX.

And that’s only one platform of the three.

I think Qri needs to do most of the key UX itself. No?

In which case, platform-native wouldn’t gain much from UX.

Platform-native could turn out to be easiest/best way anyway.

I don’t know anything of storage in `.ssh` but good to know there is something standardish.

text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+listing available private keys in a keyring may also form the basis of profile switching in the future.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+### More OS-Specific Work.
+This increases our reliance on an OS-level API that sometimes doesn't exist, and runs the risk of deviating behaviour between platforms. 
+
```
feep, line 74: Also, Apple could at any point decide that `/usr/bin/security` is a security hole and remove it or remove flags that you need, and require all apps that use keychain to be sandboxed and signed and call the API directly.

Unlikely? Yes.

But you know Apple tightens security all the time. And however unlikely, it would be a huge mess to clean up.

text/XXXX-externalize_private_keys.md
b5: I should clarify this in the RFC. Storing in OS keychain managers is only solving the problem of _"a safe place outside qri to store keys"_. The benefit of using this system is leveraging OS level access control & encryption at rest. Totally agreed we have to build our own UX for key management. If anything we _lose_ a little on the UX side, as users can delete their keys from a place outside Qri, which may cause problems.

> I don’t know anything of storage in .ssh but good to know there is something standardish.
`.ssh` isn't a standard, it's just an application with a lot of use. We've run into problems piggybacking off other software without using the binary itself for interaction in the past, so I'd like to avoid any solution that stores keys in a place intended for another program

text/XXXX-externalize_private_keys.md
b5: This would break many things. Like: all web browsers. Not a reasonable thing to be worried about

text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
+- Feature Name: Externalize Private Keys
+- Start Date: <!-- (fill me in with today's date, YYYY-MM-DD) -->
+- RFC PR: <!-- (leave this empty) -->
+- Issue: <!-- (leave this empty) -->
+
+# Summary
+[summary]: #summary
+
+Separate Qri user private key from configuration, storing in the operating system's keyring, or superceeded by an envrionrment variable `QRI_PRIVATE_KEY`.
```
dustmop, line 9: envrionrment -> environment

text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
...
+3. **remove the field** Eventually, remove the field entirely with a repo migration.
+
+In all phases the privKey field should be the _lowest_ in the loading hierarchy. Keyring values will override `Profile.PrivKey`, and the env var will override all other values.
+
+### Recovering from a deleted `.qri` folder
+Once we've externalized the private keys, we can change the behaviour of `qri setup` to first scan the keyring under the `io.qri.keys` namespace for existing keys. If a key is found, we can ask the user if they'd like to create a new account, or recover an old one.
+
+We can bolster this with calls to the default registry for the discovered public keys to add username info to the list. Profile information may not exist on the default registry, so we'll need to default to showing just the raw ProfileID. We can support this process with docs.
+
+Instead of jumping straight to building the UI for switching, we can supply a `--recover` flag that lets a user supply a ProfileID to recover from:
```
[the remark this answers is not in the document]
b5: Totally. I'd much prefer:

```
$ qri setup
We found existing Qri keys on your machine, Would you like to recover an old account, or create a new one?

▶️  create a new profile
    recover QmFjdakflonsdaEdjd (username: "feep")
    recover QmeEdjkdFTYcgFdD (username: "feepish")
```

We can use a package like [promptui](https://github.com/manifoldco/promptui) for this.

text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
...
+
+More importantly, storing private keys in configuration is a security vulnerability. Coupling sensitive information with configuration renders the entire configuration sensitive.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+We inherited this bad habit of storing private keys in configuration as a "deal with it later" approach taken from IPFS, unlike Qri, IPFS identities are not "significant" in the sense that it's not tied to an identity intended for humans. Tying an identity to a private key means we need to deal with this problem much earlier.
+
+This RFC deprecates `profile.privKey` config value, moving the the private key to the user's operating system keyring. When no keyring is available, store the keyring in a seperate file called `QRI_PRIVATE_KEY` within the `.qri` directory, and present the user with a warning that they should copy this file to a safe location.
```
dustmop, line 23: I'm a big fan of using the OS keyring if available. But I really don't think we should store them within the `.qri` as a backup. The entire initial motivation is that "Qri users are ... deleting their `.qri` directory".

Instead, we should use a user owned OS-specific directory that is normally used to store application state,
`%USERDATA%/qri_key_backup/%PROFILEID%` (on Windows) or `~/Library/qri_key_backup/${PROFILEID}` (on macos) or `~/.config/qri_key_backup/${PROFILEID}` (on other POSIX systems).

text/XXXX-externalize_private_keys.md
feep: Agree, probably not reasonable.

But you are missing a bit. They’ll never the API.

go-keyring relies on the cli (`/usr/bin/security`) to avoid calling the API.

I could write a fake VLC.app in an afternoon that prompts for a password to “install” and calls `security` to pull every password out and send them wherever. It _has_ happened in the past.

If it becomes a problem that Apple can find no workaround for, they’d pull the command without blinking.

text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
...
+
+In all cases the active private key can be overridden with an environment variable: `QRI_PRIVATE_KEY`. In all cases the key is expected to be a base64-encoded string.
```
dustmop, line 25: It would be nice to be able to support `qriswitch` still, to support multiple profiles that the user can switch between. The environment variable helps, but we should also store the users keys at multiple locations then, which is why I recommend having a subdirectory named after the `profileID` above. Using only a single file makes that much harder to do.

Also, using only a single file makes it easier to lose data if a user deletes their identity and makes a new one multiple times in a row. Keeping everything indexed by profileID avoids this problem.

text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
...
+Doing the work of generalizing secret storage can be handled by an external package: https://github.com/zalando/go-keyring. go-keyring supports Windos, OS X, and Linux. The key is stored in a `(username,password)` combination, with the "username" being the user's ProfileID (hash of public key), and the password being the base64-encoded private key. OS Keyrings support namespacing, qri's namespace will be a reversed-domain-name constant:
+
+```go
+const qriKeyringServiceID = "io.qri.keys"
+```
+
+Once setup is complete, `config.Profile.ID` stores the ProfileID string this configuration is attached to. Starting a qri process uses this as the "username" in a `Get` call:
+
+```go
+privKey, _ := keyring.Get(qriKeyringServiceID, config.Profile.ID)
```
dustmop, line 40: On some systems, using the keyring requires password entry to unlock it for the session. Do we have the ability to handle any UI needs it might have? It's probably straight-forward (hopefully handled entirely by OS's dialog prompts) just want to make sure it's been tested a bit.

text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+### Keys subsystem
+
+All of this work should happen in a new package within the qri codebase: `github.com/qri-io/qri/keys`. We need to takle key rotation next, followed by storing _many_ keys for doing access control. All of this work should happen in a new subsystem called `keys.Service`. `lib.Instance` gets a new field that points to this service.
```
dustmop, line 45: Thoughts on putting this package in `github.com/qri-io/qri/identity/keys` or `github.com/qri-io/qri/config/keys`? I worry about having too many small top-level packages.

text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Recovering from a deleted `.qri` folder
+Once we've externalized the private keys, we can change the behaviour of `qri setup` to first scan the keyring under the `io.qri.keys` namespace for existing keys. If a key is found, we can ask the user if they'd like to create a new account, or recover an old one.
+
+We can bolster this with calls to the default registry for the discovered public keys to add username info to the list. Profile information may not exist on the default registry, so we'll need to default to showing just the raw ProfileID. We can support this process with docs.
+
+Instead of jumping straight to building the UI for switching, we can supply a `--recover` flag that lets a user supply a ProfileID to recover from:
+
+```
+$ qri setup --recover QmProfileID
```
dustmop, line 64: Huge fan of this! :+1:

text/XXXX-externalize_private_keys.md
b5: big fan of this. will change.

text/XXXX-externalize_private_keys.md
b5: that's definitely a concern for me. I built a quick branch to test exactly this on OS X, it didn't ask me for a password. I'm concerned trying to list keys in a namespace would require extra privileges.

Only way to know is to test on multiple platforms.

text/XXXX-externalize_private_keys.md
Arqu: Uh, I'm a bit conflicted on this. On one hand, deeply agree that `.qri` is not the best place and we should store somewhere user/os specific. On the other hand I generally don't like obscure locations for magic configs in any app. I'm still very very much inclined to suggest that `.ssh` and it's equivalent on other OS's is the best trade off. A user dealing with that dir typically should already understand the risks with it or doesn't care either way. It's a location where people expect to have their keys added. Namespacing with a dedicated directory and file name should be plenty good + opens us up to later having multiple?. The only thing that doesn't fit is that while it is a private key it's not really an ssh key.

text/XXXX-externalize_private_keys.md
Arqu: +1 - also think the `.ssh` thing complements this further

text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
...
+### Upstream work to get recovery functionality
+The package I've suggested we build on has no high-level method for listing the keys in a namespace. We'e need to be able to do this, and haven't confirmed this will be possible across all platforms.
+
+### No Qri-specific security
+This RFC does not propose adding any key-specific security. If we used symmetric encryption on the private key itself, Qri could require a local password of it's own. In this case we're relying on the user's Operating System security architecture
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+A number of alternatives exist for externalizing secrets. We could store keys in the `.ssh` directory if it exists, and looking for guidance on
```
Arqu, line 86: I kind of went on a rumble with `.ssh` above without noticing this first. Also a point taken on the risk associated with piggybacking on `.ssh` as we have no control over it. I'm so accustomed to it that I considered it a default on any OS, which it definitely isn't. 

Back to square 1 I guess. It's just that I'd want the key in a private but _expected/known_ location.

text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
...
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+A number of alternatives exist for externalizing secrets. We could store keys in the `.ssh` directory if it exists, and looking for guidance on 
+
+# Prior art
+[prior-art]: #prior-art
+
+* Keybase - We should really do a dive on the approach keybase uses for key management before moving on to any other keypair work. _Especially_ key rotation.
```
Arqu, line 91: I like having a north-star kind of orientation on where we want to be but in this context I think containing the evolutions of keys and access control into bite sized scope is more important.

text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
...
+* Hashicorp Vault
+
+* https://github.com/keybase/go-keychain
+* https://github.com/zalando/go-keyring
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+### Loading order for configuration
+We haven't defined _when_ key loading happens in this process now that key loading is separate from configuration loading in the relation to seting up other services.
```
Arqu, line 102: I still like that the key is currently part of config as code usage wise it feels like a good home. On the other hand since we have further intentions with this and moving it into its own package, I would say it needs it's own lifecycle and ideally we should tie it to the life of the repo itself. 
Ie we are migrating, recovering, doing a key rotation - all of which should live independantly and only reflect on the repo and config. Instance loading should then follow as it does now _after_ the repo init.

text/XXXX-externalize_private_keys.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Profile Switching
+This RFC lays the groundwork for switching accounts _within_ qri, but this would have very strange results, as the data stored in `.qri` is mapped to a single user. We also don't have any UI for how this would work.
```
Arqu, line 105: I very much like this idea from the user perspective. Assuming I'm a heavy user and use it for 1 or more orgs and also personally, it gives a great way of separating contexts with users essentially working as a namespace in this scenario. This makes even more sense when we consider enterprise users/private data networks coupled with regular usage of qri on the public network.

ramfox:
> Of all the comments and inputs so far, my biggest battle is with the location of the key itself.
> 
> * Don't like the obscure location (even if it makes kind of sense) - from a user perspective it's "hard" to find
> * `.ssh` feels conceptually right, but realising this is not a given on many machines scares me completely away
> * technically the `.qri` folder makes sense namewise however we've put so much into it that we came to the current problem in the first place
> * Is a file in the home folder sensible like a `.qri_profile` (but I don't like such an important thing as a key just hanging there)
> 
> Idk, very conflicted on this.

@Arqu 
Does it help to think about the idea that this isn't the *primary* location for the key? The primary location is the os keychain which isn't obscure. The backup is just for situations where the user has removed the identity from the keychain, and can be used by qri to restore the expected behavior & would be documented. I'm not sure why the user would ever need to explicitly know where the key is stored (besides the keychain), unless they are in some super advanced use case in which case we should be documenting this all anyway and they can find it that way.

Arqu:
> > Of all the comments and inputs so far, my biggest battle is with the location of the key itself.
> > 
> > * Don't like the obscure location (even if it makes kind of sense) - from a user perspective it's "hard" to find
> > * `.ssh` feels conceptually right, but realising this is not a given on many machines scares me completely away
> > * technically the `.qri` folder makes sense namewise however we've put so much into it that we came to the current problem in the first place
> > * Is a file in the home folder sensible like a `.qri_profile` (but I don't like such an important thing as a key just hanging there)
> > 
> > Idk, very conflicted on this.
> 
> @Arqu
> Does it help to think about the idea that this isn't the _primary_ location for the key? The primary location is the os keychain which isn't obscure. The backup is just for situations where the user has removed the identity from the keychain, and can be used by qri to restore the expected behavior & would be documented. I'm not sure why the user would ever need to explicitly know where the key is stored (besides the keychain), unless they are in some super advanced use case in which case we should be documenting this all anyway and they can find it that way.

Keychain as a backup source for the "main" key location sounds good to me. Then we can keep the original in .qri and have the fallback on keystore. With that both feel like they are in their proper place.

b5:
>  I'm not sure why the user would ever need to explicitly know where the key is stored 

When they want to completely remove Qri. Having software not go away totally when being removed is always scary _\*cough\* zoom \*cough\*_.

After hearing out both sides, I feel like we should keep this RFC as-written: only place keys in `.qri` and the OS keychain, and warn the user when we can't write to the keychain. 

No matter what we do, we're going to land these changes in phases, starting with work to backup keys to the OS keychain, then adding a warning when keys aren't backed up. At that point we'll be in a situation where keys are stored in both `.qri` and the keychain, and can use that moment to evaluate if we need more work to externalize keys from the `.qri` folder.

My three main concerns with an extra folder:
* A random folder is not a secure store of sensitive information
* We're polluting users' drives even more. This gets worse when you install desktop, and reduces repo portability.
* This takes time away from a solution to the _next_ problem: key rotation and backing up identity to the registry

text/XXXX-externalize_private_keys.md
b5: Totally. To that end I could see us evolving the `.qri` folder to shard into directories of repos, with each directory being a different identity. We could use the current `.qri` contents as a "primary user".

text/XXXX-externalize_private_keys.md
b5: Yeah I think we'll always need to support both. Env var for the "master switch". 

Having directories for profiles is a very useful arrangement. I think we should consider that setup for the `.qri` directory itself.

text/XXXX-externalize_private_keys.md
b5: ok, I flip-flopped on this a bunch, and now want to leave it as-written (no extra directory) See my comment below for details.

text/XXXX-externalize_private_keys.md
feep: > On some systems, using the keyring requires password entry to unlock it for the session.

Platform keychain can require a different password on macOS, but it has to be set up that way.

But if it is only for _backing up_ the private keys I argue that it is not a concern _at all_.

feep: A different concern: if, when you create an account, the OS pops a password to unlock the keychain or a warning that “qri is trying to write to the system keychain” that might pop up some users “This seems fishy” flag.

Mr0grog: FWIW, if you want to test this in MacOS, you can open the keychain item in *Keychain Access*, click on the “Access Control” tab, remove qri from the list of apps that are always allowed access. (When an app adds a keychain item, it is automatically added to that list.) There are a whole pile of useful keychain features in MacOS that you can use (like setting a password for qri cloud as an “internet password” so Safari can also access it, or adding protocol-specific passwords, etc. etc.), but they aren’t accessible from `zalando/go-keyring`, since it’s trying to give you a reliable common approach across systems.

text/XXXX-externalize_private_keys.md
Mr0grog: > They’ll never kill the API… go-keyring relies on the cli (`/usr/bin/security`) to avoid calling the API… I could write a fake VLC.app in an afternoon that prompts for a password to “install” and calls `security` to pull every password out and send them wherever. It _has_ happened in the past.

So, just to contribute some Apple-specific context here… the same issue exists with the API. An app is no more restricted from linking the API than it is from spawning another app (e.g. `security`) and both kinds of calls are equally visible to a reviewer if you’re putting the app in the store (or equally invisible if not in the store). If your app is sandboxed, subprocesses you spawn will inherit your entitlements and restrictions. It’s always possible Apple might remove the command-line utility in the future, but it’s extraordinarily unlikely to be because of security concerns about CLI vs. API.

b5:
Ratified. Thanks all!

PR #53, opened 2020-05-03
RFC: expanded remove

b5:
Two big questions came out of the `push/pull` RFC:
* Dealing with `list` visibility
* Undoing a push

The first question is handled by the `access` RFC.

This RFC takes on "unpushing". Examining the UX of removing data from Qri (both local and network), in a way that's compatible with new APIs.


text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,192 @@
...
+
+A better UX for removing data from qri should clearly answer two common "uh oh" scenarios a user is likely to encounter:
+
+1. "I made a mistake while saving"
+2. "I want to undo a push"
+
+The first is an example of removing data from history. The second is an example of removing data from the network. In both scenarios the user is a data author, and a good answer to these questions should present well-documented, easy-to-find commands that when executed leave the user feeling that they're in a clean state. 
+
+A well-documented desctrive command should be clearly labeled as the "opposite" of a constructive command and present warnings _before_ execution. Once executed, commands remove data should show as few caveat/errors/warnings as possible, or at least connect caveats to a clear next course of action. A feeling of "clean state" comes when vommands like `qri log`, `qri list` and `qri status` match user expectations after running a destructive command.
```
ramfox, line 23: "descrtrive" => "destructive"

"commands remove data" => "commands that remove data"

b5:
> How do you remove a dataset that you have cloned? Is that part of delete, drop, or storage?

Yet another example of the RFC process working. Totally forgot about this (ahem, VERY COMMON) need. I propose we fold this functionality into `qri drop`, and have drop be about dropping local datasets as well as datasets on remotes. I'll update the RFC to indicate you'd drop a local dataset you've cloned by not specifying a remote:

```
$ qri drop  ramfox/nyc_gubernatorial_race_2019
```

This is also a chance to unify with the `--remote` flag pattern that all other commands are using, so dropping from the registry would now be:

```
$ qri drop ramfox/nyc_gubernatorial_race_2019 --remote registry
```

Attempting to drop a dataset you own will error:

```
$ qri drop me/dataset
error: cannot drop a dataset you own

because you are the author of this dataset, dropping it would remove the 
authoritative history of this dataset. Instead of dropping, you can 
delete this dataset:

  $ qri delete --all me/dataset

warning: this cannot be undone. 
```

I kinda like that local dropping takes the lead here. Considering we have more data _consumers_ than data _publishers_, and publishers should be more comfortable with qri in general.

-- --

> perhaps being prefaced by "storage" implies enough that one removes storage while the other removes... everything? 

To me that's exactly right. The meaning of `drop` is the same in both contexts (removing from persistence, not history), but with `storage`, you're saying I only want to drop dataset storage, not the `log` (which can't be dropped unless you drop _everything_ with `qri drop`).

text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,225 @@
+- Feature Name: delete and drop
+- Start Date: 2020-05-02
+- RFC PR: [#53](https://github.com/qri-io/rfcs/pulls/53)
+- Issue: <!-- (leave this empty) -->
+
+# Summary
+[summary]: #summary
+
+Rename the remove command to `delete`, and clarify it's purpose is for deleting versions from a dataset history. Create a new command `drop` for removing datasets from remotes.
```
feep, line 9: its purpose

text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,225 @@
...
+Here's the current (qri v0.9.8) description of the `remove` command:
+
+```
+$ qri remove --help
+Remove gets rid of a dataset from your qri node. After running remove, qri will
+no longer list your dataset as being available locally. 
+[rambling bit about IPFS storage redacted]
+```
+
+This text is misleading. It seems to be about storage, when the primary purpose of  `remove` is to edit history. Storage savings are a byproduct of editing one or more commits out of a history. This RFC sets aside the tight coupling between history and storage for another day, and instead focueses on fixing the role of remove by clearifying it's documentation:
```
feep, line 44: focuses
clarifying its

text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,225 @@
...
+```
+$ qri delete --help
+Delete removes commits from a dataset history. Use delete to "correct the record", 
+removing commits in a dataset history. Delete always works from the latest (HEAD) 
+commit, working backwards toward the first commit. To delete from a dataset you 
+must be able to edit it.
+
+Delete requires a "revisions" flag, specifying
+
+To delete an entire dataset and free up the name again, use `--all` (or a
```
feep, line 55: Cut the parenthetical. It may be true, but it’s not useful.

text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,225 @@
...
+
+### rename the `--drop` flag on `qri save`
+The current master branch has a `--drop` flag on the save command for "dropping components of a dataset". This RFC repurposes the `--drop` language to be about datasets you don't own, so this flag should be renamed to `--remove`. We shouldn't use `--delete`, because the word delete is now reserved for editing history.
+
+### Corner Cases
+There are a few scenarios we should make a point writing tests for:
+* Running amend on a dataset with no commits (no history) is an error
+* a drop request for a dataset name that doesn't exist locally, but _does_ exist on the remote should not error. This implies remotes should prefer local reference resolution
+* Remotes _should_ accept altered InitIDs when an owner Pushes to them. If user creates `b5/population`, pushes, then `delete --all b5/population` locally, then creates a new dataset with the same `b5/population` name, and pushes again, that should work, completely replacing the history
```
Mr0grog, line 187: FWIW, I feel like this is kind of a dangerous operation. For example, you wrote above:

> Knowing `drop` exists should make data authors more comfortable with `push`. A safe feeling that a user can `push, drop, push, drop...` should help users feel safe knowing their actions can be undone.

Obviously we’re talking about *“delete”* rather than “drop” here, but I feel like the essential thing above is that you’re suggesting someone should feel like something is safely backed up, or that there’s a safely/authoritatively stored copy somewhere else when they `push`. If they `push` an edited history, they’ve messed with the safe backup.

I think Git does actually solve this reasonably well by require a `--force` flag in this case. Qri could do the same. (Or potentially better by making the command interactive if in a TTY and `--force` is not specified?)

text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,225 @@
...
+
+```
+$ qri drop me/dataset
+error: cannot drop a dataset you own
+
+because you are the author of this dataset, dropping it would remove the 
+authoritative history of this dataset. Instead of dropping, you can 
+delete this dataset:
+
+  $ qri delete --all me/dataset
```
Mr0grog, line 164: Is there meaningful reason to worry that this isn’t quite the same thing? (At least if I’m understanding right.) Isn’t there a not-unusual (but probably not super common) scenario where I push to a registry and then drop the dataset from my machine? If I use `qri delete` for that instead, aren’t I left with a local dataset whose history diverges from the remote one? (Since it’s now a dataset in my repo but with 0 commits.)

text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,225 @@
...
+
+1. "I made a mistake while saving"
+2. "I want to undo a push"
+
+The first is an example of removing data from history. The second is an example of removing data from the network. In both scenarios the user is a data author, and a good answer to these questions should present well-documented, easy-to-find commands that when executed leave the user feeling that they're in a clean state. 
+
+A well-documented desctrive command should be clearly labeled as the "opposite" of a constructive command and present warnings _before_ execution. Once executed, commands remove data should show as few caveat/errors/warnings as possible, or at least connect caveats to a clear next course of action. A feeling of "clean state" comes when vommands like `qri log`, `qri list` and `qri status` match user expectations after running a destructive command.
+
+To improve our "undo UX", this RFC proposes three changes to qri CLI:
+1. rename `remove` to `delete`, clarify its text to talk about history instead of just storage
```
dustmop, line 26: I'm fine with this change. Though not 100% convinced it's necessary, I would lean towards thinking that `delete` is a stronger opposite to `save` than `remove`.

text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,225 @@
...
+2. Add an `amend` flag on the `save` command that _replaces_ the HEAD commit
```
dustmop, line 27: This would be great to have, awesome!

text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,225 @@
...
+3. Add a new top-level command: `drop`, for dropping datasets you don't own
```
dustmop, line 28: I'm not a fan of `drop` at all. First off, the language around "datasets you own" is tricky to explain to users. We often talk about how, before you can do certain operations, you have to get the dataset onto your machine (using what used to be `add` and what will soon become `pull`). It's confusing to talk about having a dataset in your collection vs owning it, when the difference should be transparent. Like, with a file I've downloaded versus authored, or a git repo I've cloned vs init'ed, in either case is just a thing sitting on my machine. Maybe I'm wrong on this, but all I can imagine is a user thinking "I see this dataset when I list (or see it in my Collections on Desktop) and just want to get rid of it", there should only be a single command to make it stop showing up.

This feels like reinventing the "fsi=true" problem, where qri is operating on "mode a" and "mode b", and you have to use operation A or operation B. Qri knows which operation is for each mode, but if you get it wrong there's just an error, even though qri could instead treat them as the same thing.

I understand the fear around "oops I meant to remove something I didn't own, or from a remote instead of my machine", but I would much rather have an undo escape hatch instead of a confusing UI. For instance, git has the dangerous `git reset --hard old-commit` which can be undone with `git reset --hard ORIG_HEAD` because its immutable objects exist until gc is run. We could have a similar `git restore ORIG_REF` or something thanks to logbook.

For removing from a remote I would much rather see `git remove me/oops_i_pushed_it --from-remote registry`.

text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+This text is misleading. It seems to be about storage, when the primary purpose of  `remove` is to edit history. Storage savings are a byproduct of editing one or more commits out of a history. This RFC sets aside the tight coupling between history and storage for another day, and instead focueses on fixing the role of remove by clearifying it's documentation:
+
+```
+$ qri delete --help
+Delete removes commits from a dataset history. Use delete to "correct the record", 
+removing commits in a dataset history. Delete always works from the latest (HEAD) 
+commit, working backwards toward the first commit. To delete from a dataset you 
+must be able to edit it.
```
dustmop, line 51: Generally, I'm very happy with this change; remove was not clear that it was only editing history. I would avoid repeating "remov{es/ing} commits in a dataset history" twice. And as stated above, I don't think the last sentence should be there because I don't think we should have a separate `drop` command.

text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,225 @@
...
+
+Delete requires a "revisions" flag, specifying
+
+To delete an entire dataset and free up the name again, use `--all` (or a 
+number of revisions longer than the history itself). Before doing this be sure
+to drop your dataset from other remotes
```
dustmop, line 57: We also do a lot to deal with FSI. Currently, we also have the flags `--keep-files` and `--force`. Assuming that we retain the metaphor that FSI is an incomplete future commit, it makes sense that `delete` will still have to retain these flags. For reference:

`--keep-files` deals with the fact that a user may have made modifications to the files in their working directory, which means the working directory is dirty. By default, if the directory is not clean, we refuse to remove any revisions, because that could cause work to be lost. Using this flag means "remove revisions from the repository, but don't touch the files in the working directory, I still want them in their current state".

`--force` could be better documented, but basically means "ignore any errors that would stop a remove operations, just keep going and do the best you can". This handles two main cases: 1) if the working directory has modifications, just remove the files anyway and 2) if the dataset doesn't appear to exist, there may be out-of-date state (such as logbook, or an fsi link file) so run all the steps necessary to delete what is left.

text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,225 @@
...
+
+Usage:
+  qri delete [DATASET] [--revisions|--all] [flags]
+
+Examples:
+  # delete the latest commit from annual_pop
+  $ qri delete me/annual_pop --revisions 1
+
+  # delete the latest two versions from history, keeping the underlying data
+  $ qri delete me/annual_pop --revisions 2  --keep-storage
```
dustmop, line 67: If the user doesn't use the `--keep-storage` flag, then we purge blocks from IPFS? This seem hard to do, and dangerous, and prevents any kind of "undo" functionality like what I mentioned above. It also makes us work very differently than git. I would rather that we rely on the `storage` command to run "gc" when the user wants it. Or at least, have the dangerous operation be *opt-in*, not opt-out, meaning a `--purge` flag instead.

text/XXXX-delete_and_drop.md
```diff
@@ -0,0 +1,225 @@
...
+
+Running `save` with the `--amend` flag replaces the current HEAD with a new commit. This is pulled directly from git. 
+
+Adding an `--amend` flag has the UX advantage of working language that resembles "undo" directly into the save command, which users will be familiar with before they ever need `delete`. As a git user I (b5) learned to use `--amend` before `rebase`, which I admit wasn't great, but allowed me to stick with git long enough to understand what rebase did.
+
+Users should come to think of `--amend` as a shortcut:
+
+```
+# two commands:
+$ qri delete me/dataset --revisions 1
```
dustmop, line 89: Can `--amend` only squash 1 revision? I'm fine if that's the case, just want to make sure. It certainly makes it easier to implement.

If the user has already pushed their head, do we detect that and fail with an error? Or let them implicitly branch and deal with it later somehow?

text/XXXX-delete_and_drop.md
ramfox: > you’re suggesting someone should feel like something is safely backed up, or that there’s a safely/authoritatively stored copy somewhere else when they push. If they push an edited history, they’ve messed with the safe backup.

This *is* talking about `drop` (meaning un-push aka remove from remote), not `delete`.
The quote is supposed to convey the opposite: the user should feel comfortable `push`ing, because there is an easy way to remove it from the remote if they've made a mistake.

ramfox:
from @Mr0grog 
> I kind of worry that `delete` vs. `drop` are still confusingly close terms. What if, instead of `delete`, we had a name and family of commands focused around editing the history/log?
> 
> * `qri edit`
> * `qri revise`
> * Something else?
> 
> That would also leave you with a nicer place to eventually put more complex tooling around things like squashing, rewriting older commit messages, etc.
> 
> Dunno if that’s hugely better (and it’d make the command slightly more complex, since deleting would probably be some kind of subcommand under those), but it feels at least slightly clearer and more differentiated to me.
> 
> Skimming #50, I’m also not really sure I’m clear on why there would be both `qri drop` and `qri storage drop`, since they _seem_ like they are working in the same conceptual space, and it reads like these two commands are equivalent:
> 
> ```
> > qri drop me/my_dataset
> # vs...
> > qri storage drop me/my_dataset
> ```
> 
> Or is it just that the [usage line in #50](https://github.com/qri-io/rfcs/pull/50/files#diff-885e84349f63680e8584351f782eba15R113-R114) is wrong, and it should be `qri storage drop COMMIT` (not `qri storage drop DATASET`)? Even then, it seems like these are in the same conceptual space. That `qri drop` is only for datasets as a whole and `qri storage drop` is only for parts of datasets, but never whole datasets, feels odd to me. Or I might just be misunderstanding here.

from @dustmop 
> For removing from a remote I would much rather see git remove me/oops_i_pushed_it --from-remote registry.

> If the user doesn't use the --keep-storage flag, then we purge blocks from IPFS? This seem hard to do, and dangerous, and prevents any kind of "undo" functionality like what I mentioned above. It also makes us work very differently than git. I would rather that we rely on the storage command to run "gc" when the user wants it. Or at least, have the dangerous operation be opt-in, not opt-out, meaning a --purge flag instead.


Currently `qri storage drop` is supposed to only manipulate blocks & can effect local & remote machines. `qri drop` is supposed to manipulate blocks & remove logs & can only effect remote machines. 

I agree that having so many different commands to describe similar things is confusing. Especially since `drop` in the two above contexts have very different implications: one means just remove blocks, the other means destroy logs and blocks. 

Proposing we separate out concerns based on whether they imply altering a log or not, and utilizing our newish `remote` pattern

#### qri delete
Anything with the word `delete` implies unpinning the blocks AND alters the log of a dataset.  (or this can be `remove`, tbh I don't have super strong feelings about this) 

All the previously stated local delete behaviors stay the same, with a few exceptions listed below.
But in addition to those behaviors, we have `remote` behaviors as well:
  `qri delete me/dataset --remote registry` unpins blocks & removes the log of a remote, but does not (like in our other offline&network commands) do anything locally
  `qri delete me/dataset --remote registry --revisions 1` should error: you can't manipulate the logbook of a dataset on a remote
  `qri delete not_me/dataset` *removes* the blocks and the *removes* the log of a dataset that is not in your namespace
  `qri delete not_me/dataset --revisions 1` should error: you can't manipulate the logbook of a dataset that is not yours
  If you alter your logbook locally using `qri delete`, you can update the remote log by using `qri push`

No `--keep-storage` flag, we only unpin locally 
Keep the `--keep-files` & `--force` flags
You can use the below `qri restore` command to restore a dataset in your namespace that has been removed. If you want a foreign dataset back, you need to use `qri pull`

#### qri drop
Anything with the word `drop` implies unpinning blocks ONLY. To the user this means this is their way of cleaning up storage, not effecting the actual dataset history.
  `qri drop me/dataset@version` unpins the blocks locally at that version
  `qri drop me/dataset~2` unpins the blocks locally of the two latest versions
  `qri drop me/dataset@version --remote registry` unpins the blocks at that version on that registry (again, using the remote flag implies ONLY a network action)
  `qri drop not_me/dataset@version` unpins the blocks locally of a dataset not in your namespace
To get versions back, either use the below `qri restore` command, or `qri pull` to get the dataset from a remote source


#### qri storage
This bleeds with the storage rfc, but is directly effected by this rfc

`qri storage stats me/dataset` - gives you the stats defined in the storage rfc about a dataset and its storage, locally
`qri storage stats me/dataset --remote registry` give you the stats of a dataset and its storage on the remote

`qri storage gc me/dataset` - removes the blocks from that dataset that have been unpinned
`qri storage gc me/dataset@version` removes the blocks of that version
`qri storage gc --all` removes all unpinned blocks??

no `qri storage drop` command, that is encompassed by `qri drop`

#### qri restore
And now for the command I've thought the least about that would really be the work of a future rfc.
I think it would makes sense to repurpose `qri restore` from an fsi command to a delete/drop undo command. The previous use of `qri restore` should be contained in `qri checkout`. If the fsi link already exists.

`qri restore me/dataset` - restores a dataset based on the logbook, if those blocks still exist
`qri restore me/dataset@version` - restores a dataset version if those blocks still exist

text/XXXX-delete_and_drop.md
Mr0grog: > This is talking about drop (meaning un-push aka remove from remote), not delete.

I was more referring to the commands the user is using, but I get your point that `push` then `delete` then `push` is pretty similar to `drop --remote` (except `drop --remote` wouldn’t alter what you have locally, right?)

> The quote is supposed to convey the opposite: the user should feel comfortable `push`ing, because there is an easy way to remove it from the remote if they've made a mistake.

Right. I think I was more trying to say that I don’t think the user should feel comfortable here. For my part, I’ve certainly had more than one case where `git push` failing saved my butt because I didn’t intend for my local history and the remote’s histories to have diverged. This becomes an even bigger deal when multiple machines or people are pushing to the same dataset.

Mr0grog:
> I agree that having so many different commands to describe similar things is confusing. Especially since `drop` in the two above contexts have very different implications: one means just remove blocks, the other means destroy logs and blocks.
> 
> Proposing we separate out concerns based on whether they imply altering a log or not

This is interesting, since I *don’t* see drop in these contexts being very different and I don’t think of either one as altering a log! 😜

That said, I think I was also trying to make this separation around altering the logbook. (Or the dataset’s history, which is definitely altering the log, but I’m not sure if is altering the logbook beyond adding new entries. I don’t know the logbook well enough.)

Anyway, I’m thinking of it as something like… a dataset can be in a few states on a Qri node:

1. The node doesn’t know anything about a dataset.
2. The node knows about a dataset (i.e. it has the logbook and can present the dataset name, owner, and history).
3. The node doesn’t have any of the the actual versions of the dataset (it just knows their names/IPFS addresses from the logbook). (i.e. has logbook but no blocks, I think?)
4. The node has *some* of the actual versions of the dataset, but not others (it could have any set of the versions; they don’t need to be continuous). (i.e. has logbook and all blocks of some versions?)
5. The node has all the actual versions of the dataset. (i.e. has logbook and all blocks?)

I’m thinking of `drop` as purely about editing these states (whether local or remote). `drop` doesn’t alter the logbook at all, though it could change whether I *have* the logbook.

On the other hand, `delete` (or `edit` or `revise`) would be primarily about altering what’s in the logbook, and might cause some blocks to get unpinned as a consequence of them no longer being referenced by the altered logbook.

That’s similar in a lot of ways to what you’re proposing, but:

- `qri delete me/dataset` wouldn’t really make sense as a command (instead, you want `qri drop me/dataset`). That’s a little clearer with terminology like `edit` or `revise` instead of `delete`.

- `qri delete not_me/dataset` wouldn’t even be allowed, because you can’t edit the logbook of a dataset that’s not yours/you don’t have rights to. (I hope some day Qri will support me being authorized to edit another person’s dataset, but that’s a while different story.)

- Deep agree on editing a remote’s logbook being disallowed (so `qri delete me/dataset --remote registry --revisions 1` fails; you should edit locally and `push` instead).

- I think we’re basically in agreement on `drop`, except for the two items above, where I think I’m saying what you have as `qri delete me_or_not/dataset` should be `qri drop` instead.

---

`restore` seems cool! Wondering if `recover` would be clearer for what you’re proposing, though.

Likewise, `qri storage gc me/dataset` and `qri storage gc me/dataset@version` would be very nice if they’re not absurdly complex to implement (since I’m not sure IPFS has a way to GC just the DAG under a given block).

Mr0grog:
Enh, got up this morning and not sure `recover` is really any better/clearer a name than `restore`. Too nuanced a difference, if any.

text/XXXX-delete_and_drop.md
b5: yes amend can only ever squash one commit.


The `amend` will be marked in logbook. When pushing after amending the receiving remote will update and remove the orphaned commit, or at least know they _can_ remove the orphaned commit. 

They don't have the option to branch, but we could later build tools that show more detailed histories, allowing a user to see dropped commits still recorded in logbook.

dustmop:
I feel as is the logbook should purely be an implementation detail, and the user shouldn't have to know about it in order to understand the differences between commands. Details about logbook should never reach the UI layer.

Mr0grog:
I agree! I only included the logbook in my explanation to connect it to implementation. But was trying to frame `drop` around the question “what data do I have?” And `delete`/`edit`/`revise` around “changing what the history of this dataset is.”

b5:
Thanks so much everyone for your feedback. Clearly lots of room for improvement here, and I'm delighted to see everyone dig in to work this one out.

To me it's clear the original sin of this RFC introducing two words for removing things. Without any additional info, it's _way_ too hard to know these do different things:

```
$ qri delete user/dataset
$ qri drop user/dataset
```

We _must_ have an easy, top-level command that takes stuff away, and success here is defined as the command for taking stuff working with as few caveats as possible. I'm inclined to back @ramfox's proposal almost wholesale, including moving `drop` into the storage command. That would leave us with:

```
$ qri delete user/dataset
$ qri storage drop user/dataset@/ipfs/QmFoo
```

I very much agree we (the designers of qri) should be cognizant of when a user is taking an action that will edit history. @ramfox's proposal addresses these concerns by ensuring the API guards against unintended consequences. Presenting clear errors will be a requirement. I'm a big fan of incorporating a `--force` flag to slow down users.

The upside of this approach is a user can only see one word for going backwards (`delete`), this also leaves more top-level space for the suite of commands @Mr0grog is proposing. I very much agree commands like `revise` are in our future, but we can't write meaningful RFCs against that world yet because we haven't picked a collaboration model.

-- --

> Anyway, I’m thinking of it as something like… a dataset can be in a few states on a Qri node:

This is a great way of thinking about things @Mr0grog, I'd re-write your set of states like this:

1. The node doesn’t know anything about a dataset.
2. The node has a _resolved reference_ to a dataset. It's established the `username, datasetname, logbook identifier, and head reference` from a trusted source, but that's it.
3. The node has the logbook.
4. The node has the logbook + a subset of versions 
5. The node has the logbook + all versions listed in the logbook

> I’m thinking of `drop` as purely about editing these states (whether local or remote). drop doesn’t alter the logbook at all, though it could change whether I have the logbook.

This is why I think we should move `drop` into storage. Ideally users will almost _never_ need `$qri storage drop`, once we get some sort of `$ qri storage gc` command in place

Arqu:
Oh my, lot's of stuff here. Going through it seems like a lot of concerns are already addressed.

My only remaining concern is one of the first brought up: `drop` and `delete` being too close in meaning while both being descrutcive, doing an accidental `drop` feels much worse.

`delete` being a `commit` level command I'd rather call it something like `revert` where we just unroll back in history. Or `reset` like `git` does it?

Also very much into the idea of `--force` if overriding instead of appending history on push.

b5:
OK. I've rewritten this entire RFC to incorporate as much feedback as possible. `drop` is dropped. What's left is an expansion and clarification of how remove should work to address common user concerns of deleting stuff.

I've kept the original `remove` term for a few reasons:
* No one felt `delete` was an improvement
* This RFC is no longer a breaking change
* delete is already an alias for the remove command in qri v0.9.8

b5:
I'm concerned about review fatigue, so I've asked @dustmop to take a close look as a point-person for making sure this new RFC is headed in the right direction

text/0031-expanded_remove.md
```diff
@@ -0,0 +1,236 @@
...
+$ qri remove not_me/dataset --revisions 1
+ERROR
+```
+should error: you can't manipulate the logbook of a read-only dataset.
+
+If you alter your logbook locally using qri remove, you can update the remote log by using qri push.
+
+
+### Use the logbook
+Delete actions should examine the history of a logbook to present warnings to a user that help them "clean up" when deleteing stuff. As an example: `$ qri remove --all` on a dataset should look for `push` operations and warn user they should `drop` from a list of remotes.
```
dustmop, line 173: deleteing -> deleting

Question: does this warning abort the `remove` operation? I would suggest it does, but can be ignored with `--force`, since that's how other "errors during remove" currently operate.

text/0031-expanded_remove.md
```diff
@@ -0,0 +1,236 @@
...
+
+### rename the `--drop` flag on `qri save`
+The current master branch has a `--drop` flag on the save command for "dropping components of a dataset". This RFC repurposes the `--drop` language to be about datasets you don't own, so this flag should be renamed to `--remove`. We shouldn't use `--delete`, because the word delete is now reserved for editing history.
```
dustmop, line 176: Does this still apply? I think we can keep using `save --drop [component]` now, and I think I'm more fond of that than changing the flag to `save --remove [component]`.

text/0031-expanded_remove.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Corner Cases
+There are a few scenarios we should make a point writing tests for:
+* Running amend on a dataset with no commits (no history) is an error
+* a drop request for a dataset name that doesn't exist locally, but _does_ exist on the remote should not error. This implies remotes should prefer local reference resolution
+* Remotes _should_ accept altered InitIDs when an owner Pushes to them. If user creates `b5/population`, pushes, then `delete --all b5/population` locally, then creates a new dataset with the same `b5/population` name, and pushes again, that should work, completely replacing the history
```
dustmop, line 182: I would rephrase this as "Remotes should uniquely identify datasets by their initID, rather than their name, when a new dataset is pushed to them. A name can be changed or reused while a initID can not." The important part here is that, while we ensure names are unique, they can change, so they aren't stable ways to identify datasets. The way it is currently phrased inverts the relationship, making it seem like a *name* lasts forever, and the initID for that name is being altered.

Rather, we should think of it like a user signing up for a new website. Though each user name can only be used once, it shouldn't be the primary key that is used internally to refer to their database row. That way the username can be changed later on, or reused if the first user deletes their account.

text/0031-expanded_remove.md
```diff
@@ -0,0 +1,236 @@
...
+$ git push origin --delete feature/login
+```
+
+We do something similar now with `$ qri publish --unpublish ` now, and it's a viable alternative to having dedicated commands. Looking at the registry, users seem to be finding & using the `--unpublish` flag, which is heartening. My concern is qri wants to deviate from git by giving users fine-grained control over version storage. Layering deletes atop push operations leaves little room for these detailed needs.
+
+
+# Prior art
+[prior-art]: #prior-art
+
+Heroku's `rollback` command plays an important UX inspiration for `drop`. Having a top level "go back" command is part of what makes working with heroku feel safe.
```
dustmop, line 211: drop -> `remove --remote`

text/0031-expanded_remove.md
```diff
@@ -0,0 +1,236 @@
...
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+### Storage management
+This RFC puts storage management out of scope.
+
+
+### logbook garbage collection
+Running `qri remove --all` drops a bunch a storage and "caps off" a dataset history with a delete operation, preventing the dataset from showing up in list operations. Delete doesn't remove remove logbook data. We'll have to figure out a way to clean up the logbook itself with a plubming command of some sort.
```
dustmop, line 221: plubming -> plumbing

text/0031-expanded_remove.md
```diff
@@ -0,0 +1,236 @@
...
+
+Here's the proposed help text for the new remove command:
+
+```
+$ qri remove --help
+Remove deletes datasets from qri.
+
+For read-only datasets you've pulled from others, Remove gets rid of a dataset 
+from your qri node. After running remove, qri will no longer list your dataset 
+as being available locally, and will free up storage space.
```
dustmop, line 70: Not sure if this is leftover from a previous revision, or new functionality, or just worded strangely. If I remove a dataset I've cloned from elsewhere, then run `remove`, are those blocks immediately freed up? Perhaps you meant, "and may eventually free up the storage space." Otherwise this seems both difficult to implement, and best left to the `storage` command.

text/0031-expanded_remove.md
```diff
@@ -0,0 +1,236 @@
...
+The remote flag can only be used to completely remove a dataset from a remote.
+To edit history on a remote, run delete locally and use `qri push` to send the
+updated history to the remote. Any command run with the remote flag has no
+effects on local data.
+
+Usage:
+  qri remove DATASET [DATASET...] [flags]
+
+Examples:
+  # delete a dataset
```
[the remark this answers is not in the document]
b5: shouldn't work in read-write cases. I'll update the text

text/0031-expanded_remove.md
b5: agree it's confusing, and will change to your proposed language.

text/0031-expanded_remove.md
b5: agreed. removed

text/0031-expanded_remove.md
b5: totally. adding in

text/0031-expanded_remove.md
```diff
@@ -0,0 +1,236 @@
...
+All the previously stated local delete behaviors stay the same, with a few exceptions listed below.
+But in addition to those behaviors, we have remote behaviors as well:
+```
+$ qri remove me/dataset --remote registry
+```
+Unpins all blocks & removes the log on the registry because the registry has a read-only copy. Does nothing locally.
+
+Attempting to drop revisions from a read-only dataset replica is an error:
+```
+$ qri remove me/dataset --remote registry --revisions 1
```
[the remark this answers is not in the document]
b5: > Maybe this is something we solve later on once collaboration is defined, but it seems like the authorship requirement of logbook may be getting in the way of the preferable UX. Luckily this behavior doesn't work against changing it in the future.

Very much agree here. I see this as a stepping stone to that UX. first we need to figure out how to do delegated PKI.

b5:
Ok, this is ready for a final sign-off

b5:
Approved! It's in! Huge thanks to everyone for digging in on this one. Qri's going to be much better for it 😄

PR #54, opened 2020-05-03
RFC: move publish into a new access command

b5:
Two big questions came out of the push/pull RFC:

* Dealing with list visibility
* Undoing a push
The second question is handled by the [delete and drop RFC](https://github.com/qri-io/rfcs/pull/53)

This RFC takes on @ramfox's question in review of `push/pull` on publication status, setting the stage for real-deal access control.

### DIFF
--- a/text/0032-access_command.md
+++ b/text/0032-access_command.md
@@ -6,7 +6,7 @@
 # Summary
 [summary]: #summary
 
-Decouple dataset visibility from data pushing, moving `publish` & `--unpublish` into a new subcommand: `access`. Position `access` for future commands.
+Decouple dataset visibility from data pushing, moving `publish` & `--unpublish` into a new subcommand: `access`. Position `access` for future commands like access control and encryption. Rename `publish/unpublish` to `visible/hidden`.
 
 # Motivation
 [motivation]: #motivation
@@ -28,6 +28,7 @@ The actual process of moving data around is the subject of another RFC on pushin
 This RFC proposes four concrete changes:
 * create a new `access` subcommand to view & edit access control for users & datasets
 * move `publish` and `unpublish` into access, changing both commands to edit _only_ `list` visibility (not push data)
+* rename `publish/unpublish` to `visibile/hidden`.
 * add an `info` command under access to show access info for a dataset
 * pushing to the registry automatically runs "publish". dropping from the registry automatically unpublishes
 
@@ -47,49 +48,49 @@ Usage:
 
 Available Commands:
   info           show access control details for a dataset
-  publish        list dataset for others to see and pull
-  unpublish      unlist dataset, removing it from public visibility
+  hidden         unlist dataset, removing it from public visibility
+  visible        list dataset for others to see and pull
 ```
 
-Publish used to have a note talking about how it pushes data. This RFC assumes that work is now handled by `qri push`, so now publish only controls `list` visibility:
+The old `publish` used to have a note talking about how it pushes data. This RFC assumes that work is now handled by `qri push`. the new `visible` command only controls listing & feed visibility:
 
 ```
-$ qri access publish --help
-Publish makes your dataset publically visible to others. While online, peers 
+$ qri access visible --help
+Visible makes your dataset publically visible to others. While online, peers 
 that connect to you can only see datasets and versions that you've published. 
-Publishing a dataset always makes all previous history entries available, and 
-any updates to a published dataset will be immediately visible. Published 
+Making a datset a dataset always makes all previous history entries available, 
+and any updates to a visible dataset will be immediately visible. Published 
 datasets will show up in feeds.
 
-New datasets are *not* published by default, remaining unlisted until published.
+New datasets are *not* visible by default, and remain hidden until pushed.
 
 Pushing to the registry requires a dataset be either encrypted or published.
-By default pushing a dataset to the registry also publishes.
+By default pushing a dataset to the registry also makes a dataset visible.
 
 Usage:
-  qri access publish [DATASET] [DATASET...] [flags]
+  qri access visible [DATASET] [DATASET...] [flags]
 
 Examples:
-  # Publish a dataset so others can see it with "qri list":
-  $ qri access publish me/dataset
+  # make a dataset visible so others can see it with "qri list":
+  $ qri access visible me/dataset
 
-  # Publish a few datasets:
-  $ qri access publish me/dataset me/other_dataset
+  # make a few datasets visible at once:
+  $ qri access visible me/dataset me/other_dataset
 ```
 
-Unbpublish used to be a flag. In this RFC it's bumped up to a subcommand of it's own, to match other reciprocal commands like `push/pull`. 
+Unbpublish used to be a flag. In this RFC it's bumped up to a subcommand of its own, to match other reciprocal commands like `push/pull`. Here's it's renamed to `hidden`, and again it only deals with weather a dataset will show up in listing & feeds. the `hidden` command does _not_ move data around.
 
 ```
-$ qri access unpublish --help
-Unpublish "unlists" a dataset. An unlisted dataset will not show up when
-browsing dataset lists or feeds. Others can still access an unpublished dataset
-if they know it's name.
+$ qri access hidden --help
+hidden "unlists" a dataset. A hidden dataset will not show up when browsing 
+dataset lists or feeds. Others can still access a hidden dataset if they know
+its name.
 
-Unpublished datasets cannot be pushed to the registry, but can be pushed to
-peers that accept unpublished datasets from you.
+hidden datasets cannot be pushed to the registry, but can be pushed to peers 
+that accept hidden datasets from you.
 
 Usage:
-  qri access unpublish [DATASET] [DATASET...] [flags]
+  qri access hidden [DATASET] [DATASET...] [flags]
 
 Examples:
   # unpublish a dataset, removing it from any lists or feeds
@@ -103,7 +104,7 @@ Flags:
 ```
 
 
-Publish & Unpublish on their own may not seem like they warrant being moved into a subcommand. To understand the justification for the `access`, this is how `access` _might_ look one day, after we've landed encryption and role-based access control (RBAC) for datasets:
+`visible` & `hidden` on their own may not seem like they warrant being moved into a subcommand. To understand the justification for the `access`, this is how `access` _might_ look one day, after we've landed encryption and role-based access control (RBAC) for datasets:
 
 ```
 $ qri access --help
@@ -114,8 +115,8 @@ Usage:
 
 Available Commands:
   info           show access control details for a dataset
-  publish        list dataset for others to see and pull
-  unpublish      unlist dataset, removing it from public visibility
+  visible        list dataset for others to see and pull
+  hidden         unlist dataset, removing it from public visibility
   encrypt        restrict access to a dataset with a secret key
   decrypt        remove access restrictions from a dataset
   allow          give permissions to a peer for an encrypted dataset
@@ -135,11 +136,11 @@ An example output for a published dataset:
 ```
 $ qri access info me/dataset
 can-edit:  true
-published: true
+visible:   true
 
 You can edit this dataset.
 
-This dataset is "published". Other users may see it when you're online, and if
+This dataset is visible. Other users may see it when you're online, and if
 pushed to the registry it will show up in feeds for other users to browse and
 pull.
 ```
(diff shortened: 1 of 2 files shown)


text/XXXX-access_command.md
```diff
@@ -0,0 +1,213 @@
...
+  $ qri access publish me/dataset me/other_dataset
+```
+
+Unbpublish used to be a flag. In this RFC it's bumped up to a subcommand of it's own, to match other reciprocal commands like `push/pull`. 
+
+```
+$ qri access unpublish --help
+Unpublish reverses the publish process, removing a dataset from public view. 
+Unpublishing a dataset will try to "clean up", asking remotes this dataset has
+been published to
```
ramfox, line 86: unfinished sentence

text/XXXX-access_command.md
```diff
@@ -0,0 +1,213 @@
...
+
+Unpublished datasets cannot be pushed to the registry, but can be pushed to
+peers that accept unpublished datasets from you.
+
+Usage:
+  qri access unpublish [DATASET] [DATASET...] [flags]
+
+Examples:
+  # unpublish a dataset, removing
```
ramfox, line 95: unfinished thought

"removing it from the registry"?

text/XXXX-access_command.md
```diff
@@ -0,0 +1,213 @@
...
+This dataset is "published". Other users may see it when you're online, and if
+pushed to the registry it will show up in feeds for other users to browse and
+pull.
+```
+
+`$ qri access info me/dataset` for now will only but it will be _very_ helpful for viewing dataset permsissions that control both who can access a dataset and what they can do with it.
+
+### Publish on registry push
+Pushing unencrypted datasets† to the registry _requires_ a dataset be published. To keep the old behaviour of a one-liner publish, pushing to the registry will automatically run publish on the user's behalf. The push command will present this as user feedback:
+
```
ramfox, line 165: What does this imply for remotes?
Can you push a dataset to a remote but not have it published?
If it's not published and the remote only allows published datasets, is it the remote's responsibility to notify the user?

ramfox (commented):
I do think having more the one dimension to describe the availability of a dataset is good.

Now we can talk about if it's published or not and, separately, where it has been pushed.

My major questions relate to remotes in all this and how the behavior is different or the same then the registry. Assuming this behavior can change per remote depending on settings, & that a remote can accept unpublished or published datasets from users & will not list unpublished datasets.

text/0032-access_command.md
```diff
@@ -0,0 +1,215 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Access control is really irritating.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+### `listed` and `unlisted`
+Instead of the publish/unpublish terminology, we could use a more direct reference to list visibility. We can't use `list` directly because `qri access list` should be the command for listing all access-controlled datasets (if such a thing turns out to be necessary). An alternative would be to use a different conjucation or tense for list like "listed" & "unlisted". using a past tense is the best I could come up with, and in my opinion `$ qri access listed` isn't very memorable, and feels like your asking about things that have been listed in the past?
```
feep, line 200: Strong dislike. The term is not as good.

But even if it was the best term, any command or flag that runs that close to another command should always be avoided.

text/0032-access_command.md
```diff
@@ -0,0 +1,215 @@
...
+
+The other thing to keep in mind: unlisted datasets shouldn't show up in feeds. The registry requires "published: true", but feeds built on the peer-2-peer network won't. Peers building feeds should only use datasets that are published. This clearly aligns
```
feep, line 202: aligns...

unfinished

text/0032-access_command.md
```diff
@@ -0,0 +1,215 @@
...
+pull.
+```
+
+An example for a dataset you _don't_ own:
+```
+$ qri access info ca-state-parks/park-features
+can-edit:  false
+published: true
+
+You cannot edit this dataset, attempting to save to it's history will create a
```
feep, line 153: its

text/0032-access_command.md
```diff
@@ -0,0 +1,215 @@
...
+
+  # Publish a few datasets:
+  $ qri access unpublish me/dataset me/other_dataset
+ 
+Flags:
+  -r, --remove     issue "remove all" commands to any published remotes
+```
+
+
+Publish & Unpublish on their own may not seem like they warrant being moved into a subcommand. To understand the justification for the `access`, this is how `access` _might_ look one day, after we've landed encryption and role-based access control (RBAC) for datasets:
```
feep, line 106: I did not like the name _access_ until I got to here.

Added comment at the top to clarify in summary.

text/0032-access_command.md
```diff
@@ -0,0 +1,215 @@
...
+  $ qri access publish me/dataset me/other_dataset
+```
+
+Unbpublish used to be a flag. In this RFC it's bumped up to a subcommand of it's own, to match other reciprocal commands like `push/pull`. 
+
+```
+$ qri access unpublish --help
+Unpublish "unlists" a dataset. An unlisted dataset will not show up when
+browsing dataset lists or feeds. Others can still access an unpublished dataset
+if they know it's name.
```
feep, line 86: its

text/0032-access_command.md
```diff
@@ -0,0 +1,215 @@
...
+
+Examples:
+  # Publish a dataset so others can see it with "qri list":
+  $ qri access publish me/dataset
+
+  # Publish a few datasets:
+  $ qri access publish me/dataset me/other_dataset
+```
+
+Unbpublish used to be a flag. In this RFC it's bumped up to a subcommand of it's own, to match other reciprocal commands like `push/pull`.
```
feep, line 80: of its own
the other ‘it's’ is correct.

text/0032-access_command.md
```diff
@@ -0,0 +1,215 @@
+- Feature Name: access_command
+- Start Date: 2020-05-02
+- RFC PR: [#54](https://github.com/qri-io/rfcs/pulls/54)
+- Issue: <!-- (leave this empty) -->
+
+# Summary
+[summary]: #summary
+
+Decouple dataset visibility from data pushing, moving `publish` & `--unpublish` into a new subcommand: `access`. Position `access` for future commands.
```
feep, line 9: ...for future access-control commands.
or permissions, rights, something.

Something to clarify that it will be for more than visibility.

text/0032-access_command.md
dustmop: Agree here, "listed" is not great, but I think "published" is also ambiguous, especially in the case of working within a network of known peers. How about "visible" vs "hidden", or "public" vs "secret", something along those lines?

b5: big fan of "visible" & "hidden"! I'll update the RFC to use those terms instead.

The reason for not "public" and "secret": collides with the statement "all datasets in qri are public".

Arqu (approved):
So, I've went through the proposal a couple times with questions poping up and resolving but leaving me in a bit of a confused state. I agree with the overal RFC but want to clarify a couple things:

- publish/unpublish should essentially say this dataset is now publicly (in)visible
- it should absolutely not move any data around
- I don't think that pushing to a public registry should automatically mean it's also published it can be available but unlisted. Once published all subsequent pushing should also be public because access control says it's a published dataset (this has the caveat that the user might expect the other scenario but I'd rather have users see a message clarifying this than doing `push` which results in publicly listed dataset that they might not want)
- The above would also mean that we have to have a `push` flag for "unlisted" pushing since otherwise users cannot push without having the dataset public
- My reasoning would be: being able to have the repo live somewhere else while it's WIP in case I run over my laptop with a car in the meantime (this obviously depends on the fact we can rotate keys and recover users anyways)

I'm kind of left and right on this but if it were to be implemented as stated, would be fine.

text/0032-access_command.md
```diff
@@ -0,0 +1,215 @@
...
+Remotes, on the other hand _can_ accept unpublished datasets. Remote configuration will get a `requirePublish` configuration parameter to get the same registry-like behaviour. Pushing unpublished datasets is aimed at the collaborator use-case, where a datasets are explicitly pushed to trusted colleagues. when a user is pushed an unpublished dataset, it will show up when they run `qri list` locally, but this dataset won't be broadcast back to the network.
+
+† _encrypted datasets don't exist yet._
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### The `Published` field
+For some time we've modelled `published` as a boolean field outside the dataset itself, stored as a logbook operation. We've struggled with this field not capturing the problem of _where_ a dataset's been published. Re-conceiving of `published` as a "permission bit" clarifies the intent of this field. Published now answers "can the existence of this dataset be broadcasted" for the _entire history_ of a dataset.
```
Mr0grog, line 182: > We've struggled with this field not capturing the problem of _where_ a dataset's been published.

I [asked some questions around this](https://github.com/qri-io/rfcs/pull/49#discussion_r410790390) over on the push/pull PR, but I’m curious about the *where* part here, and am not sure how this proposal addresses that (assuming I understand future directions).

Specifically, there’s been talk in the past about a future where there are many registries, and of course we have the situation where you could push to many peers today. If `published` is going to be modeled as a “permission bit,” is it doing so on a particular (peer/remote/registry + dataset) combination, or just on the dataset? Does marking it as published mark it as published *everywhere*? If not, it’s unclear in here how how mark it as published in all the various locations it might be. (I’m also unclear as to whether qri tracks the various peers/remotes a dataset might have been pushed to, other than the one registry you have configured for all datasets in your repo.) Or are we always just talking about the registry your have currently configured?

text/0032-access_command.md
```diff
@@ -0,0 +1,215 @@
...
+
+* Google Docs Sharing settings
+* Unlisted Vimeo Videos
+* RBAC - Role Based Access Control
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+### An "Access" Data Structure
+This RFC describes `published` as a _permissions flag_. At some point we should be looking to build a data structure that describes permissions. `Published` might be a field on that data structure. The challenge here: permissions will apply to, what, users? groups? Needs more thought.
```
Mr0grog, line 215: Or is this the eventual answer to my question above? 🤔

text/0032-access_command.md
```diff
@@ -0,0 +1,224 @@
...
+  visible        list dataset for others to see and pull
+```
+
+The old `publish` used to have a note talking about how it pushes data. This RFC assumes that work is now handled by `qri push`. the new `visible` command only controls listing & feed visibility:
+
+```
+$ qri access visible --help
+Visible makes your dataset publically visible to others. While online, peers 
+that connect to you can only see datasets and versions that you've published. 
+Making a datset a dataset always makes all previous history entries available,
```
dustmop, line 61: "Making a datset a dataset" -> "Making a dataset "public""?

text/0032-access_command.md
```diff
@@ -0,0 +1,224 @@
...
+  hidden         unlist dataset, removing it from public visibility
+  visible        list dataset for others to see and pull
+```
+
+The old `publish` used to have a note talking about how it pushes data. This RFC assumes that work is now handled by `qri push`. the new `visible` command only controls listing & feed visibility:
+
+```
+$ qri access visible --help
+Visible makes your dataset publically visible to others. While online, peers 
+that connect to you can only see datasets and versions that you've published.
```
dustmop, line 60: This paragraph is still talking about datasets being "published".

text/0032-access_command.md
```diff
@@ -0,0 +1,224 @@
...
+This RFC proposes permissions like visibility _only apply to datasets_. If you `push` a visible dataset to remote A, and peer B downloads that dataset, it's also `visible` for peer B.
+
+Later when we add access control, if you say "this dataset can _only_ be accessed by peer B", that will apply to anyone with the dataset.
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### The `Visible` field
+For some time we've modelled `published` as a boolean field outside the dataset itself, stored as a logbook operation. We've struggled with this field not capturing the problem of _where_ a dataset's been published. Re-conceiving the new `visible` field as a "permission bit" clarifies the intent of this field. Published now answers "can the existence of this dataset be broadcasted" for the _entire history_ of a dataset.
```
dustmop, line 191: We're still storing changes to visibility using logbook entries, right?

text/0032-access_command.md
```diff
@@ -0,0 +1,224 @@
...
+
+This dataset is visible. Other users may see it when you're online, and if
+pushed to the registry it will show up in feeds for other users to browse and
+pull.
+```
+
+`$ qri access info me/dataset` for now will only but it will be _very_ helpful for viewing dataset permsissions that control both who can access a dataset and what they can do with it.
+
+### Visible on registry push
+Pushing unencrypted datasets† to the registry _requires_ a dataset be visible. To keep the old behaviour of a one-liner publish, pushing to the registry will automatically run `$ qri access visible` on the user's behalf. The push command will present this as user feedback:
```
dustmop, line 165: Discussed earlier on video, I agree that we should change this language to say "Pushing unencrypted datasets† to the registry defaults to making that dataset become visible."

We should support pushing hidden datasets, but it should require additional flags from the user, making it an opt-in

dustmop (commented):
Almost there! Just a few comments

ramfox (approved):
Specifically IN2 this w/ visible/hidden vocab change

text/0032-access_command.md
b5: > If published is going to be modeled as a “permission bit,” is it doing so on a particular (peer/remote/registry + dataset) combination, or just on the dataset?

Just on the dataset. Marking a dataset as published (we're now calling it "visible") is the same anywhere a dataset is pushed. If you don't want a user to have a hidden dataset, don't push it to them.

> I’m also unclear as to whether qri tracks the various peers/remotes a dataset might have been pushed to, other than the one registry you have configured for all datasets in your repo

Yep! We track this by adding a "push" operation to the dataset oplog that also stores the multiaddress that was pushed to. This is the same for remotes or the registry (wherever possible we model the registry as just another remote). 

Upside of this: you can get hints about where a dataset is replicated by looking at logbook data.

There's work to do here to _optimistically add the operation to the pushed logbook_, and rollback on failure but it's a technical detail.

text/0032-access_command.md
b5: I don't think it is. In my mind this permissions structure is independent of where a thing is replicated.

text/0032-access_command.md
b5: yes. I'll add a note

b5:
Ok, I've incorporated feedback, and re-worked the interaction between `visible` and `push`. There's a new consequence that comes from having `visible` be tied to the dataset. From the updated RFC:

> Without setting a dataset to visible everyone (including the user) will not be able to see the pushed dataset in list operations on the remote after pushing._
> To solve this, pushing an unencrypted dataset to any remote makes it visible before pushing by default. This keeps the current (v0.9.8) behaviour of a one-liner publish, pushing to the registry will automatically run `$ qri access visible` on the user's behalf. The push command will present this as user feedback.

To push a hidden dataset, this update proposes adding a `--hidden` flag to `push` that will keep the dataset hidden _and_ present a warning to the user so they know what `--hidden` is doing, and won't be confused when the dataset doesn't show up for them in a list operation.

If others are ok with this change, I think this one is good-to-go, and would love a sign off 😄

dustmop (approved):
Looks good! Some small spelling cleanups, and one question that is non-blocking.

text/0032-access_command.md
```diff
@@ -173,26 +182,29 @@ pushing 28 versions of chriswhong/some_dataset...
 X of XX blocks pushed for version 12 August 2019 (Qmss4h552h3j2h33)...
 ```
 
-A `--hidden` flag on `push` overrides this behaviour, which will present a warning:
+Defaulting to `visible` avoids a potential UX problem when a dataset is pushed: without setting a dataset to visible everyone (including the user) will not be able to see the pushed dataset in list operations on the remote after pushing. Users who want to push hidden datasets should first make them hidden with `$ qri access hidden`, then push. Pushing a hidden dataset should present a warning:
```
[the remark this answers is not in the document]
b5: I think the long term solution is driving a "who is asking" parameter deeper into remote behavior. We'll _need_ this as a part of access control, but it could also remove the need for the "you're pushing a hidden dataset" warning, because remotes can show users lists contextualized to their identities.

Because the registry is a remote, it would inherit this behavior

b5:
AMAZING. RATIFIED. Thanks everyone for your feedback!

